PR #142, opened 2022-07-21
chore: upgrade antique dependencies

kvcache:
There are some material changes in io.grpc netty methods that don't
work well with the older versions in here. No reason not to use the
newer versions we've tested on the service.

Integration tests pass locally against a momento endpoint.


bruuuuuuuce (approved):
:shipit:

gautamomento (approved):
Thank you

PR #147, opened 2022-09-23
chore: fix otel dependencies

kvcache:
we don't depend on anything other than otel apis in our implementation.
we should not have such a wide runtimeDependency closure as to take an
opinion about (e.g.) the exporter


bruuuuuuuce (approved):
:shipit:

cprice404 (approved):

gautamomento (approved):
❤️

PR #150, opened 2022-11-17
chore: adds code scanning workflow

eaddingtonwhite:


cprice404 (commented):
looks like this one needs some more config.  Also, if we are going to run these we might need to add something to CONTRIBUTING.md that tells devs how to repro locally if failures are encountered?

eaddingtonwhite:
@cprice404 good call out. I made a follow-up issue for this on our internal tracking board. Will come back and do consistently on all SDK repos.

tylerburdsall (approved):

PR #151, opened 2022-11-17
chore: adds dep review

eaddingtonwhite:


cprice404 (commented):
seems like this one might need additional config

eaddingtonwhite:
@cprice404 I think this one is gtg now that fixed readme and auth tokens. Code scanning one does indeed need some work still but dep check I think is GTG.

cprice404 (approved):

PR #155, opened 2023-03-07
chore: update getCache response

nand4011:
Update the response returned by the get and getAsync methods to match the pattern we have established in other SDKs. The response has three implementations: Hit, Miss, and Error, and it no longer throws. It is intended to be used with pattern matching.

Change the deleteAsync method to delete instead of getting.

Add a test dependency on AssertJ, a neat fluent assertion library.

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
@@ -1,10 +1,10 @@
 package momento.sdk;
 
+import static momento.sdk.TestUtils.randomString;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.*;
 
 import java.time.Duration;
-import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import momento.sdk.exceptions.AuthenticationException;
 import momento.sdk.exceptions.InternalServerException;
@@ -61,10 +61,10 @@ final class SimpleCacheClientTest extends BaseTestClass {
 
   @Test
   public void createCacheGetSetDeleteValuesAndDeleteCache() {
-    final String cacheName = UUID.randomUUID().toString();
-    final String alternateCacheName = UUID.randomUUID().toString();
-    final String key = UUID.randomUUID().toString();
-    final String value = UUID.randomUUID().toString();
+    final String cacheName = randomString("name");
+    final String alternateCacheName = randomString("alternateName");
+    final String key = randomString("key");
+    final String value = randomString("value");
 
     target.createCache(cacheName);
     target.createCache(alternateCacheName);
@@ -89,10 +89,10 @@ final class SimpleCacheClientTest extends BaseTestClass {
 
   @Test
   public void shouldFlushCacheContents() {
-    String cacheName = UUID.randomUUID().toString();
-    String key = UUID.randomUUID().toString();
-    String value = UUID.randomUUID().toString();
-    long ttl1HourInSeconds = Duration.ofHours(1).getSeconds();
+    final String cacheName = randomString("name");
+    final String key = randomString("key");
+    final String value = randomString("value");
+    final long ttl1HourInSeconds = Duration.ofHours(1).getSeconds();
 
     target.createCache(cacheName);
     try {
@@ -136,22 +136,21 @@ final class SimpleCacheClientTest extends BaseTestClass {
       // Unable to hit control plane
       final InternalServerException e =
           assertThrows(
-              InternalServerException.class,
-              () -> client.createCache(UUID.randomUUID().toString()));
+              InternalServerException.class, () -> client.createCache(randomString("cacheName")));
       assertTrue(e.getMessage().contains("Unable to reach request endpoint."));
 
       // But gets a valid response from Data plane
       final CacheGetResponse getResponse = client.get("helloCache", "key");
       assertThat(getResponse).isInstanceOf(CacheGetResponse.Error.class);
-      assertThat(((CacheGetResponse.Error) getResponse).exception())
-          .isInstanceOf(AuthenticationException.class);
+      assertThat(((CacheGetResponse.Error) getResponse))
+          .hasCauseInstanceOf(AuthenticationException.class);
 
       assertThrows(AuthenticationException.class, () -> client.set("helloCache", "key", "value"));
 
       final CacheGetResponse asyncGetResponse = client.getAsync("helloCache", "key").join();
       assertThat(asyncGetResponse).isInstanceOf(CacheGetResponse.Error.class);
-      assertThat(((CacheGetResponse.Error) asyncGetResponse).exception())
-          .isInstanceOf(AuthenticationException.class);
+      assertThat(((CacheGetResponse.Error) asyncGetResponse))
+          .hasCauseInstanceOf(AuthenticationException.class);
 
       final ExecutionException setException =
           assertThrows(
@@ -167,20 +166,20 @@ final class SimpleCacheClientTest extends BaseTestClass {
 
       // Can reach control plane.
       assertThrows(
-          AuthenticationException.class, () -> client.createCache(UUID.randomUUID().toString()));
+          AuthenticationException.class, () -> client.createCache(randomString("cacheName")));
 
       // Unable to reach data plane
       final CacheGetResponse getResponse = client.get("helloCache", "key");
       assertThat(getResponse).isInstanceOf(CacheGetResponse.Error.class);
-      assertThat(((CacheGetResponse.Error) getResponse).exception())
-          .isInstanceOf(InternalServerException.class);
+      assertThat(((CacheGetResponse.Error) getResponse))
+          .hasCauseInstanceOf(InternalServerException.class);
 
       assertThrows(InternalServerException.class, () -> client.set("helloCache", "key", "value"));
 
       final CacheGetResponse response = client.getAsync("helloCache", "key").join();
       assertThat(response).isInstanceOf(CacheGetResponse.Error.class);
-      assertThat(((CacheGetResponse.Error) response).exception())
-          .isInstanceOf(InternalServerException.class)
+      assertThat(((CacheGetResponse.Error) response))
+          .hasCauseInstanceOf(InternalServerException.class)
           .hasMessageContaining("Unable to reach request endpoint.");
 
       final ExecutionException setException =
--- a/momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
@@ -1,8 +1,7 @@
 package momento.sdk;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.*;
 
 import java.time.Duration;
 import java.util.UUID;
@@ -12,8 +11,6 @@ import momento.sdk.exceptions.InternalServerException;
 import momento.sdk.exceptions.InvalidArgumentException;
 import momento.sdk.exceptions.NotFoundException;
 import momento.sdk.messages.CacheGetResponse;
-import momento.sdk.messages.CacheGetStatus;
-import momento.sdk.messages.CacheSetResponse;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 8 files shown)


momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
```diff
@@ -129,24 +131,29 @@ public void throwsExceptionWhenClientUsesNegativeDefaultTtl() {
 
   @Test
   public void initializesSdkAndCanHitDataPlaneForUnreachableControlPlane() {
-    try (SimpleCacheClient client =
+    try (final SimpleCacheClient client =
         SimpleCacheClient.builder(BAD_CONTROL_PLANE_JWT, DEFAULT_TTL_SECONDS).build()) {
       // Unable to hit control plane
-      InternalServerException e =
+      final InternalServerException e =
```
cprice404, line 137: ❤️ finals

momento-sdk/src/main/java/momento/sdk/SimpleCacheClient.java
```diff
@@ -10,16 +10,7 @@
...
-import momento.sdk.messages.CacheGetResponse;
-import momento.sdk.messages.CacheSetResponse;
-import momento.sdk.messages.CreateCacheResponse;
-import momento.sdk.messages.CreateSigningKeyResponse;
-import momento.sdk.messages.DeleteCacheResponse;
-import momento.sdk.messages.FlushCacheResponse;
-import momento.sdk.messages.ListCachesResponse;
-import momento.sdk.messages.ListSigningKeysResponse;
-import momento.sdk.messages.RevokeSigningKeyResponse;
+import momento.sdk.messages.*;
```
cprice404, line 13: nit - we tend to try to avoid wildcard imports.

momento-sdk/src/main/java/momento/sdk/SimpleCacheClient.java
```diff
@@ -168,10 +159,6 @@ public ListSigningKeysResponse listSigningKeys(Optional<String> nextToken) {
    * @param key The key to get
    * @return {@link CacheGetResponse} containing the status of the get operation and the associated
    *     value data.
-   * @throws momento.sdk.exceptions.PermissionDeniedException
-   * @throws ClientSdkException if key is null
-   * @throws NotFoundException
-   * @throws momento.sdk.exceptions.InternalServerException
    */
   public CacheGetResponse get(String cacheName, String key) {
     return scsDataClient.get(cacheName, key);
```
cprice404, line 164: Note: in the dotnet SDK when we were ramping up for 1.0 release, we decided to entirely remove the synchronous/blocking API.  I think we should consider this for Java too; I'm leaning towards that approach.  I think it would be helpful to have a few things in the examples dir that show how to block on the future from an async call to demonstrate that users can use it in a synchronous fashion without too much ceremony if they like, but beyond that I think we should probably just rip out all of the synchronous signatures.

If you agree with this, I would suggest floating that idea for feedback in #engineering.

momento-sdk/src/main/java/momento/sdk/SimpleCacheClient.java
```diff
@@ -400,8 +375,8 @@ public CompletableFuture<CacheDeleteResponse> deleteAsync(String cacheName, Stri
    * @throws NotFoundException
    * @throws momento.sdk.exceptions.InternalServerException
    */
-  public CompletableFuture<CacheGetResponse> deleteAsync(String cacheName, byte[] key) {
-    return scsDataClient.getAsync(cacheName, key);
```
cprice404, line 404: lol

momento-sdk/src/main/java/momento/sdk/messages/CacheGetResponse.java
```diff
@@ -1,109 +1,80 @@
...
-   */
-  public Optional<byte[]> byteArray() {
-    if (status != CacheGetStatus.HIT) {
-      return Optional.empty();
+    /**
+     * Value stored in the cache as a byte array.
+     *
+     * @return Value stored for the given key.
+     */
+    public byte[] byteArray() {
```
cprice404, line 25: we've established a convention that all of our value accessors will have a name that starts with the word `value` so that it's easy to find them across different response types, and so that code completion will always group them together nicely.

cprice404 (commented):
This is looking good to me so far.  I left a handful of comments.

The other thing that would be helpful is to put up a separate WIP PR that shows how the example code will change with this response pattern.  The tests tend to look quite a bit different than actual production consumption.

Once you have some updated examples we can ping a few folks that have Java backgrounds for more input before we fully commit and start moving forward.

momento-sdk/src/main/java/momento/sdk/SimpleCacheClient.java
nand4011: I need to change my intellij settings to avoid that.

momento-sdk/src/main/java/momento/sdk/SimpleCacheClient.java
nand4011: I wouldn't mind removing the synchronous methods.

cprice404: okay let's start a thread about this in #engineering today.  if we decide to do it, it can be a separate PR obvs.

cprice404:
Cool, this lgtm now, but let's get up a second PR that shows what the example consumption of this would look like vs. the old way.  Once we have that I can solicit a little more input from a few other folks and then we should be good to move forward.

cprice404 (commented):
lgtm now.  left a comment about adding a PR to update examples so we can socialize it, and then we can move forward from there.

momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
```diff
@@ -64,25 +61,30 @@ void teardown() {
 
   @Test
   public void createCacheGetSetDeleteValuesAndDeleteCache() {
-    String cacheName = UUID.randomUUID().toString();
-    String key = UUID.randomUUID().toString();
-    String value = UUID.randomUUID().toString();
+    final String cacheName = UUID.randomUUID().toString();
```
malandis, line 64: As this will be very common, let's extract "randomUUIDString" to a utility method.

momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
```diff
@@ -64,25 +61,30 @@ void teardown() {
...
-    assertEquals(CacheGetStatus.MISS, getAfterDeleteResponse.status());
+      final CacheGetResponse getResponse = target.get(cacheName, key);
+      assertThat(getResponse).isInstanceOf(CacheGetResponse.Hit.class);
+      assertThat(((CacheGetResponse.Hit) getResponse).valueString()).isEqualTo(value);
 
-    CacheGetResponse getForKeyInSomeOtherCache = target.get(System.getenv("TEST_CACHE_NAME"), key);
-    assertEquals(CacheGetStatus.MISS, getForKeyInSomeOtherCache.status());
+      target.delete(cacheName, key);
+      final CacheGetResponse getAfterDeleteResponse = target.get(cacheName, key);
+      assertThat(getAfterDeleteResponse).isInstanceOf(CacheGetResponse.Miss.class);
```
malandis, line 80: In other test suites, we include a string explaining what the value was when the assertion fails. Will the test runner output this by default?

momento-sdk/src/intTest/java/momento/sdk/SimpleCacheDataPlaneBlockingTest.java
```diff
@@ -106,48 +102,64 @@ void itemDroppedAfterTtlExpiresWithTracing() throws Exception {
...
 
   @Test
   public void setGetDeleteWithByteKeyValuesMustSucceed() {
-    byte[] key = {0x01, 0x02, 0x03, 0x04};
-    byte[] value = {0x05, 0x06, 0x07, 0x08};
-    SimpleCacheClient cache =
-        SimpleCacheClient.builder(authToken, DEFAULT_ITEM_TTL_SECONDS).build();
-    cache.set(cacheName, key, value, 60);
+    final byte[] key = {0x01, 0x02, 0x03, 0x04};
+    final byte[] value = {0x05, 0x06, 0x07, 0x08};
```
malandis, line 138: I recommend generating a UUID as a byte array (similar to the way you did for strings). That way if we run the tests in quick succession we won't collide with past test data.

momento-sdk/src/main/java/momento/sdk/messages/CacheGetResponse.java
```diff
@@ -1,109 +1,80 @@
...
-    if (status != CacheGetStatus.HIT) {
-      return Optional.empty();
+    /**
+     * Gets the retrieved value as a {@link ByteBuffer}.
+     *
+     * @return the value.
+     */
+    public ByteBuffer valueByteBuffer() {
+      return value.asReadOnlyByteBuffer();
     }
```
malandis, line 36: Have we landed on needing both of these?

momento-sdk/src/main/java/momento/sdk/messages/CacheGetResponse.java
```diff
@@ -1,109 +1,80 @@
...
-  public Optional<InputStream> inputStream() {
-    if (status != CacheGetStatus.HIT) {
-      return Optional.empty();
+    /**
+     * Gets the retrieved value as an {@link InputStream}.
+     *
+     * @return the value.
+     */
+    public InputStream valueInputStream() {
+      return value.newInput();
```
malandis, line 63: Same with this one, ie do we want the minimum set of value accessors to: utf-8 string, other encoded string, byte array, byte buffer, input stream? If not relevant for this PR, ignore this :-)

malandis (commented):
I really appreciate the thorough tests. Just a couple questions and a suggestion.

momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
nand4011: There will be no description by default other than saying the type is wrong. I could add descriptions.

momento-sdk/src/main/java/momento/sdk/messages/CacheGetResponse.java
nand4011: I also don't know if we really need all of those methods in the response. I just moved over the existing ones. @cprice404 what are your feelings on having a batteries included response like this vs one that just returns a byte[] and string?

nand4011:
@malandis I didn't make large changes to the tests other than where needed due to the response object change. I agree with your comments but I think I would rather do a test cleanup pr or make changes in future prs when changing more of the responses.

momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
kvcache: and that utility method should take a non-optional prefix

nand4011: Done.

momento-sdk/src/intTest/java/momento/sdk/SimpleCacheDataPlaneBlockingTest.java
nand4011: This class is being deleted in the next pr so I'm leaving this for now.

momento-sdk/src/main/java/momento/sdk/messages/CacheGetResponse.java
cprice404: huge +1 from me on trimming these down.  we need string, and we need one bytes accessor and i'd stop there.

the question is, for the bytes one, which type should we expose?  tbh i'm tempted to choose `inpustream`.

@kvcache  any thoughts?

kvcache: Our GetResponse (and all payload objects) are fully materialized: I vote we should give users byte[] and String (and String only because we have taken on the utf-8 symmetric functionality).

Hiding their `byte[]` behind an InputStream will only annoy someone who is using bytes APIs (hello protocol buffers), and it won't matter at all to someone who is using nio streams (trivial to wrap a byte[] in your stream type of choice).

We could possibly in the future support huge items for which we would vend batches of bytes at a time. In that possible future, I think returning an InputStream would be the most truthful and friendly type available.

nand4011: Updated.

momento-sdk/src/main/java/momento/sdk/messages/CacheGetResponse.java
```diff
@@ -26,41 +23,13 @@ public byte[] valueByteArray() {
...
     }
 
-    /**
-     * Gets the retrieved value as a {@link ByteBuffer}.
-     *
-     * @return the value.
-     */
-    public ByteBuffer valueByteBuffer() {
-      return value.asReadOnlyByteBuffer();
-    }
```
kvcache: This one is the only one I'll miss _a little_. It could be nice to have a zero-copy way to get the bytes, but I'm +1 on waiting for users to express an opinion on it!

cprice404 (approved):
here we go

PR #156, opened 2023-03-08
chore: create example to show new response pattern matching

nand4011:
Temporarily modify the example to show how the new response style can be used with pattern matching. DO NOT MERGE.

### DIFF
--- a/examples/lib/src/main/java/momento/client/example/MomentoCacheApplication.java
+++ b/examples/lib/src/main/java/momento/client/example/MomentoCacheApplication.java
@@ -29,10 +29,33 @@ public class MomentoCacheApplication {
 
       System.out.println(String.format("Getting value for key=`%s`", KEY));
 
-      CacheGetResponse getResponse = simpleCacheClient.get(CACHE_NAME, KEY);
-      System.out.println(String.format("Lookup resulted in: `%s`", getResponse.status()));
-      System.out.println(
-          String.format("Looked up value=`%s`", getResponse.string().orElse("NOT FOUND")));
+      { // Java 8 style
+        final CacheGetResponse getResponse = simpleCacheClient.getAsync(CACHE_NAME, KEY).join();
+        if (getResponse instanceof CacheGetResponse.Hit) {
+          final CacheGetResponse.Hit hit = (CacheGetResponse.Hit) getResponse;
+          System.out.println("Hit! Found value: " + hit.valueString());
+        } else if (getResponse instanceof CacheGetResponse.Miss) {
+          System.out.println("Miss!");
+        } else if (getResponse instanceof CacheGetResponse.Error) {
+          final CacheGetResponse.Error error = (CacheGetResponse.Error) getResponse;
+          System.out.println("Error encountered: " + error.exception().toString());
+        } else {
+          System.out.println("Unknown response type.");
+        }
+      }
+
+      { // Java 14 style
+        final CacheGetResponse getResponse = simpleCacheClient.getAsync(CACHE_NAME, KEY).join();
+        if (getResponse instanceof CacheGetResponse.Hit hit) {
+          System.out.println("Hit! Found value: " + hit.valueString());
+        } else if (getResponse instanceof CacheGetResponse.Miss) {
+          System.out.println("Miss!");
+        } else if (getResponse instanceof CacheGetResponse.Error error) {
+          System.out.println("Error encountered: " + error.exception().toString());
+        } else {
+          System.out.println("Unknown response type.");
+        }
+      }
     }
     printEndBanner();
   }
--- a/examples/lib/build.gradle.kts
+++ b/examples/lib/build.gradle.kts
@@ -19,11 +19,15 @@ repositories {
 }
 
 dependencies {
-    implementation("momento.sandbox:momento-sdk:0.18.0")
+    // This is gross but most of gradle's cross-project dependency stuff doesn't work since examples isn't
+    // a sub-module of the sdk project.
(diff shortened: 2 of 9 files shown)


examples/lib/src/main/java/momento/client/example/MomentoCacheApplication.java
```diff
@@ -29,10 +29,33 @@ public static void main(String[] args) {
...
+        if (getResponse instanceof CacheGetResponse.Hit hit) {
+          System.out.println("Hit! Found value: " + hit.valueString());
+        } else if (getResponse instanceof CacheGetResponse.Miss) {
+          System.out.println("Miss!");
+        } else if (getResponse instanceof CacheGetResponse.Error error) {
+          System.out.println("Error encountered: " + error.exception().toString());
+        } else {
+          System.out.println("Unknown response type.");
+        }
+      }
```
cprice404, line 58: Thanks @nand4011 !

@kvcache @malandis @pgautier404 @poppoerika wdyt of this?  It looks pretty much identical to the way we did it in dotnet so it seems like a reasonable path forward?

kvcache: I think this is state of the art until jep420 lands in (probably) v19 at which point this should be updated to:

```
switch (getResponse) {
    case CacheGetResponse.Hit hit -> System.out.println("Hit! Found value: " + hit.valueString());
    case CacheGetResponse.Miss _  -> System.out.println("Miss!");
    default -> System.out.println("Unknown response type.");
}
```

Can we still return Error? Should that be removed?

cprice404: why wouldn't we be returning Error?

cprice404: `Unavailable` etc.?

cprice404: also this says Jep420 is available as a preview in 17/18?  Might be worth including examples of that too then when we flesh out the examples @nand4011 

https://openjdk.org/jeps/420

kvcache: wouldn't that end up an exception at join()? https://github.com/momentohq/client-sdk-java/blob/main/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java#L247-L248

kvcache: I'm probably forgetting or missing something.

cprice404: lol @kvcache you wrote the blog post that instigated this philosophy of surfacing errors are part of the return type instead of throwing exceptions.... I'm confused :)

nand4011: The error handling isn't all done yet but I figured with an error response I would capture all the ways the method could throw as an error.

kvcache: I know 😄 I think I'm just missing where we actually make it a friendly return value. Don't mind me, this example looks nice.

cprice404: this is just a sketch/proposal of what it will look like after the changes.

kvcache: ahh that helps me understand, thanks.

nand4011: In the parent pull request that logic is changed to return an error instead of throwing an exception: [https://github.com/momentohq/client-sdk-java/blob/31dc780424c21e7b1527e570f4ad2b0750924d89/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java#L255](url)

examples/lib/src/main/java/momento/client/example/MomentoCacheApplication.java
```diff
@@ -29,10 +29,33 @@ public static void main(String[] args) {
...
+
+      { // Java 14 style
+        final CacheGetResponse getResponse = simpleCacheClient.getAsync(CACHE_NAME, KEY).join();
+        if (getResponse instanceof CacheGetResponse.Hit hit) {
+          System.out.println("Hit! Found value: " + hit.valueString());
+        } else if (getResponse instanceof CacheGetResponse.Miss) {
+          System.out.println("Miss!");
+        } else if (getResponse instanceof CacheGetResponse.Error error) {
+          System.out.println("Error encountered: " + error.exception().toString());
```
kvcache, line 54: What would you think of making CacheGetResponse.Error extend RuntimeException so people could directly throw the error if they wanted to? (minimize annoyance for throwy people) Is there something aside from `.exception()` that you'd hang on this?

nand4011: I like that idea. The class is already structured like an exception already with a cause. I wouldn't be opposed to it.

examples/lib/src/main/java/momento/client/example/MomentoCacheApplication.java
poppoerika: > Thanks @nand4011 !
> 
> @kvcache @malandis @pgautier404 @poppoerika wdyt of this? It looks pretty much identical to the way we did it in dotnet so it seems like a reasonable path forward?

It looks good to me! This patter is pretty much identical to what we have in the .NET SDK.
Thank you for getting it started, @nand4011 🙇‍♀️

PR #157, opened 2023-03-08
chore: remove synchronous methods.

nand4011:
Remove the synchronous versions of the cache methods. The async versions can easily be made synchronous, and we reduce the code surface significantly.


cprice404 (approved):
it begins

cprice404 (approved):

PR #160, opened 2023-03-10
feat: convert responses to new pattern part 1

nand4011:
Convert CacheDeleteResponse, CacheSetResponse, CreateCacheResponse, DeleteCacheResponse, and FlushCacheResponse to use the new inheritance pattern.


momento-sdk/src/main/java/momento/sdk/ValidationUtils.java
```diff
@@ -4,6 +4,7 @@
 
 // Should rely on server for all validations. However, there are some that cannot be delegated and
 // instead fail in grpc client, like providing null inputs or a negative ttl.
+// TODO: validation should not let exceptions escape from cache methods
```
nand4011, line 7: This is here because validation will currently throw exceptions from the new methods even though that is incorrect. I didn't want to fix it until I had all the responses switched over. That pr will immediately follow the last response conversion pr.

cprice404: this is fine for now, but you should also discuss this with @pgautier404 and @malandis before you do the work to change the behavior.  For some types of exceptions that are just straight up user bugs, we do actually throw exceptions.  I think Schwern wrotes some notes about it in one of our standards-and-practices docs.

cprice404 (dismissed):
looking good, thanks!

cprice404 (approved):

momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
```diff
@@ -136,30 +145,22 @@ public void initializesSdkAndCanHitDataPlaneForUnreachableControlPlane() {
     try (final SimpleCacheClient client =
         SimpleCacheClient.builder(BAD_CONTROL_PLANE_JWT, DEFAULT_TTL_SECONDS).build()) {
       // Unable to hit control plane
-      final InternalServerException e =
-          assertThrows(
-              InternalServerException.class, () -> client.createCache(randomString("cacheName")));
-      assertTrue(e.getMessage().contains("Unable to reach request endpoint."));
+      final CreateCacheResponse createResponse = client.createCache(randomString("cacheName"));
+      assertThat(createResponse).isInstanceOf(CreateCacheResponse.Error.class);
+      assertThat(((CreateCacheResponse.Error) createResponse))
```
malandis, line 150: No automatic type narrowing?

malandis (commented):
Is the error revamp part of this? I don't see other fields like grpc error code in `SdkException`. Also we will need `toString()` implementations for each of the responses.

momento-sdk/src/intTest/java/momento/sdk/SimpleCacheClientTest.java
nand4011: I haven't done any error handling changes yet other than converting throws to returning an error type. i will do a separate comprehensive error handling pass.

PR #162, opened 2023-03-10
chore: improve error handling for new response types

nand4011:
Add try/catch blocks to all the cache methods to capture any validation errors. The cache methods should return an error result instead of throwing.


cprice404 (commented):
let's bottom out on the whole "to throw or not to throw" question before we merge this.  lmk if you need help finding the info.

PR #163, opened 2023-03-15
feat: update error handling

nand4011:
Update the error handling to match other SDKs. Errors that occur during client creation should throw an exception. Errors that occur during a client call should never throw and should return an error response.

Add internal grpc error details to the exceptions.

Add toString() methods to the responses that contain data. Empty responses and error responses can use the default toString().

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/exceptions/WrappedSdkException.java
+++ b/momento-sdk/src/main/java/momento/sdk/exceptions/WrappedSdkException.java
@@ -1,0 +1,8 @@
+package momento.sdk.exceptions;
+
+/** An SdkException that is a wrapper around another SdkException. */
+public class WrappedSdkException extends SdkException {
+  public WrappedSdkException(SdkException cause) {
+    super(cause.getMessage(), cause, cause.getTransportErrorDetails().orElse(null));
+  }
+}
--- a/momento-sdk/src/main/java/momento/sdk/exceptions/WrappedSdkException.java
+++ b/momento-sdk/src/main/java/momento/sdk/exceptions/WrappedSdkException.java
@@ -1,8 +1,0 @@
-package momento.sdk.exceptions;
-
-/** An SdkException that is a wrapper around another SdkException. */
-public class WrappedSdkException extends SdkException {
-  public WrappedSdkException(SdkException cause) {
-    super(cause.getMessage(), cause, cause.getTransportErrorDetails().orElse(null));
-  }
-}
--- a/momento-sdk/src/main/java/momento/sdk/messages/CacheDeleteResponse.java
+++ b/momento-sdk/src/main/java/momento/sdk/messages/CacheDeleteResponse.java
@@ -1,6 +1,7 @@
 package momento.sdk.messages;
 
 import momento.sdk.exceptions.SdkException;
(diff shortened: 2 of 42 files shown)


momento-sdk/src/main/java/momento/sdk/exceptions/WrappedSdkException.java
```diff
@@ -0,0 +1,8 @@
+package momento.sdk.exceptions;
+
+/** An SdkException that is a wrapper around another SdkException. */
+public class WrappedSdkException extends SdkException {
+  public WrappedSdkException(SdkException cause) {
```
cprice404, line 5: nit: couldn't you just add this as an alternate constructor for the main SdkException class?

momento-sdk/src/main/java/momento/sdk/messages/CacheDeleteResponse.java
```diff
@@ -13,10 +14,10 @@ class Success implements CacheDeleteResponse {}
    * thrown, or the cause of the error can be retrieved with {@link #getCause()}. The message is a
    * copy of the message of the cause.
    */
-  class Error extends SdkException implements CacheDeleteResponse {
+  class Error extends WrappedSdkException implements CacheDeleteResponse {
```
cprice404, line 18: nit: I'm not 100% sure where we have landed on this across all of the SDKs, but I had been advocating for exposing a property/method called `errorCode` on these, which would provide access to the actual enum of all of our different error types.  That way people could write switch statements that key off of the enum rather than having to do type checking on the exception object. See:

https://github.com/momentohq/standards-and-practices/blob/main/docs/client-specifications/error-handling.md#error-types

idk if we've stuck with this through all of the SDKs so I'm open to discussion on it, but that was my original hope/intent.

@pgautier404 it might be good for you to review this PR and weigh in on state-of-the-art.

cprice404 (commented):
looking good, just a few questions

momento-sdk/src/main/java/momento/sdk/messages/CacheDeleteResponse.java
pgautier404: Yes, we should definitely be using an errorCode property to match against a public enum of the available codes listed in that standards and practices doc. You might take a look at [the nodejs implementation](https://github.com/momentohq/client-sdk-nodejs/blob/main/src/errors/errors.ts), which I think makes the pattern pretty clear. Let me know if it would be useful to take a look at this and/or implementations in other SDKs together, @nand4011.

momento-sdk/src/main/java/momento/sdk/exceptions/WrappedSdkException.java
nand4011: Well, that would be too easy. 🤦 I'll switch to that.

momento-sdk/src/main/java/momento/sdk/messages/CacheDeleteResponse.java
nand4011: I didn't implement the error codes because they are 1:1 with exceptions. I have no problem adding them in though.

cprice404 (approved):
error

PR #164, opened 2023-03-15
chore: make dependencies explicit

nand4011:
Add the dependency analyze plugin to fail the build if there are used undeclared dependencies.

Make all undeclared dependencies found with the plugin explicit.

Remove the apache commons dependency since it was only used in one place.

Replace the messages module with a direct dependency on client-protos.


nand4011:
Converting this to a draft because the protos work might change some things.

messages/build.gradle.kts
```diff
@@ -10,8 +10,11 @@ dependencies {
 
     api("io.grpc:grpc-protobuf:${rootProject.ext["grpcVersion"]}")
     api("io.grpc:grpc-stub:${rootProject.ext["grpcVersion"]}")
-    api("com.google.protobuf:protobuf-java-util:${rootProject.ext["protobufVersion"]}")
+    api("com.google.protobuf:protobuf-java:${rootProject.ext["protobufVersion"]}")
     compileOnly("org.apache.tomcat:annotations-api:6.0.53") // necessary for Java 9+
+    implementation("com.google.guava:guava:${rootProject.ext["guavaVersion"]}")
```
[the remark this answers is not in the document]
cprice404: not sure I understand the question here, but it sounds like you are on pause on this one while you sort out the client protos release process.  ping me again when you're ready and we can discuss!

cprice404 (commented):
sounds like we should sync after the client-protos packaging is sorted a bit

messages/build.gradle.kts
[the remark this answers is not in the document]
nand4011: My confusion about submodule dependencies is now academic as we can remove the submodle in favor of a real dependency.

CONTRIBUTING.md
```diff
@@ -4,8 +4,6 @@
     * `brew install openjdk`
 1. Clone repo
     * `git clone git@github.com:momentohq/client-sdk-java.git`
-1. Initialize git submodule
-    * `git submodule init && git submodule sync && git submodule update`
```
cprice404, line 8: these are among the best two deleted lines of code ever

build.gradle.kts
```diff
@@ -1,14 +1,20 @@
 ext["grpcVersion"] = "1.47.0"
+ext["guavaVersion"] = "31.0.1-android" // This version is in sync with grpcVersion
```
cprice404, line 2: wth?  do we really want to pin to an android version?

cprice404 (approved):
lgtm, one question about the android version of the dep but not a blocker

build.gradle.kts
nand4011: That is the version used internally by the grpc libraries and the one we were transitively using before. It is the recommended version to use if you want android and JRE compatibility, as per [https://github.com/google/guava/wiki/ReleasePolicy#flavors](url). I don't mind changing it, but not using it could be a problem if someone comes up with a use case for momento directly in an android app.

cprice404: nah just surprising but your explanation makes sense.

PR #177, opened 2023-03-22
chore: update sdk dependencies

nand4011:
Move dependency versions into a gradle catalog in libs.versions.toml.

Switch to the newest version of client-protos.

Switch jjwt-jackson to jjwt-gson because we already have a dependency on gson.

Update misc dependencies.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,0 +1,35 @@
+[versions]
+assertj = "3.24.2"
+grpc = "1.53.0"
+protobuf = "3.22.2"
+jjwt = "0.11.5"
+junit = "5.9.2"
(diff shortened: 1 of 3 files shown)


gradle/libs.versions.toml
nand4011, line 1: This is the fancy new way to declare dependency versions in gradle. I am not attached to doing it this way, so if there are any objections I can go back to the previous way.

cprice404: let's try it out!

cprice404: @kvcache you might be interested in knowing that This Is A Thing, for other projects

cprice404 (approved):

PR #179, opened 2023-03-24
feat: make project publishable on sonatype

nand4011:
Remove artifactory publishing functionality.

Add momento.publishable-java-lib.gradle.kts containing the configuration to publish to maven central. This could be modified to support multiple modules in the future.

Change the cd workflow to publish to sonatype.

Update workflow action versions.


.github/workflows/cd.yml
```diff
@@ -1,35 +1,28 @@
...
+      uses: actions/checkout@v3
       with:
         token: ${{ secrets.MOMENTO_MACHINE_USER_GITHUB_TOKEN }}
-        submodules: true
+
     - name: Set up JDK 16
-      uses: actions/setup-java@v2
+      uses: actions/setup-java@v3
       with:
         java-version: 16
```
nand4011, line 19: Is there a particular reason we aren't using an LTS version of java for this? I would prefer to switch to 17 at least.

.github/workflows/cd.yml
```diff
@@ -1,35 +1,28 @@
...
-        distribution: 'adopt'
-    - name: Grant execute permission for gradlew
-      run: chmod +x gradlew
-    - name: Build with Gradle
-      run: ./gradlew clean build
-#    - name: Run integration tests
-#      run: ./gradlew integrationTest
+        distribution: 'coretto'
```
nand4011, line 20: Since AdoptOpenJdk is no longer developed I changed to coretto. I am open to any of them, however.

.github/workflows/cd.yml
```diff
@@ -1,35 +1,28 @@
...
-    - uses: actions/checkout@v2
+    - name: Checkout project
+      uses: actions/checkout@v3
       with:
         token: ${{ secrets.MOMENTO_MACHINE_USER_GITHUB_TOKEN }}
-        submodules: true
-    - name: Set up JDK 16
-      uses: actions/setup-java@v2
+
+    - name: Set up JDK 17
```
nand4011, line 16: I updated to Java 17 since 16 wasn't an lts release. I could try 20 but that is only 3 days old.
I switched to corretto since AdoptOpenJdk isn't developed anymore. We can use any version supported by the setup java plugin if we have a strong preference. [https://github.com/actions/setup-java#supported-distributions](url)

.github/workflows/cd.yml
```diff
@@ -1,35 +1,28 @@
...
+      uses: actions/setup-java@v3
+      with:
+        java-version: 17
+        distribution: 'corretto'
```
nand4011, line 20: I switched to corretto since AdoptOpenJdk isn't developed anymore. We can use any version supported by the setup java plugin if we have a strong preference. [https://github.com/actions/setup-java#supported-distributions](url)

.github/workflows/cd.yml
cprice404: nope 17 is good

cprice404 (approved):
lgtm.  before we change this to auto-release we need to set up a "manual release" job and move the publish call over to that, but it doesn't matter to me whether you want to do that now-ish or come back to it after we get some of the collections in.

PR #182, opened 2023-03-24
chore: Remove member variable optionals and fix warnings

nand4011:
Remove member variable and method argument optionals. Java optionals are sad compared to Kotlin's and provide no null guarantees, so it is valid to supply a method that takes an optional with a null.

Add a jsr305 dependency to get access to @Nonnull and @Nullable annotations.

Annotate the cache client constructor chain with @Nullable and @Nonnull. These are the only places where the sdk can throw an exception if nulls are provided.

Fix a few misc warnings.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/CacheClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/CacheClient.java
@@ -6,6 +6,8 @@ import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import momento.sdk.messages.CacheDeleteResponse;
 import momento.sdk.messages.CacheGetResponse;
 import momento.sdk.messages.CacheIncrementResponse;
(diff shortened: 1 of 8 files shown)


momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -25,20 +27,16 @@ public final class CacheClient implements Closeable {
   private final ScsDataClient scsDataClient;
 
   CacheClient(
-      String authToken,
-      Duration itemDefaultTtl,
-      Optional<OpenTelemetry> telemetryOptional,
-      Optional<Duration> requestTimeout) {
+      @Nonnull String authToken,
+      @Nonnull Duration itemDefaultTtl,
+      @Nullable OpenTelemetry openTelemetry,
```
cprice404, line 32: I thought we got rid of this otel stuff?

cprice404 (approved):
looks fine.  I am rusty on my java but I presume we are keeping Optionals if there are any public accessors where the value might be null?

also you can get rid of the rest of that otel stuff whenever you have a chance

momento-sdk/src/main/java/momento/sdk/CacheClient.java
nand4011: I must have misunderstood you and removed the tests. I can put up a separate pr that removes the rest of it.

nand4011:
> looks fine. I am rusty on my java but I presume we are keeping Optionals if there are any public accessors where the value might be null?

Yeah. Methods that return something that might be null should return an optional.

rishtigupta (approved):

rishtigupta (dismissed):

rishtigupta (approved):

PR #187, opened 2023-03-28
feat: Add List Concatenate Back API support

rishtigupta:


momento-sdk/src/main/java/momento/sdk/ValidationUtils.java
```diff
@@ -12,6 +12,9 @@ final class ValidationUtils {
   static final String A_NON_NULL_KEY_IS_REQUIRED = "A non-null key is required.";
   static final String A_NON_NULL_VALUE_IS_REQUIRED = "A non-null value is required.";
   static final String CACHE_NAME_IS_REQUIRED = "Cache name is required.";
+  static final String LIST_NAME_CANNOT_BE_NULL = "Cache name cannot be null.";
```
nand4011, line 15: Should be list instead of cache

nand4011 (approved):

PR #188, opened 2023-03-28
chore: Add ListFetch API support

rishtigupta:
## PR Description
- Add ListFetch API
- Add corresponding integ tests

## Build

## Issue
#170

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
@@ -4,6 +4,7 @@ import static io.grpc.Metadata.ASCII_STRING_MARSHALLER;
 import static java.time.Instant.now;
 import static momento.sdk.ValidationUtils.checkCacheNameValid;
 import static momento.sdk.ValidationUtils.checkListNameValid;
+import static momento.sdk.ValidationUtils.checkListSliceStartEndValid;
 import static momento.sdk.ValidationUtils.checkSetNameValid;
 import static momento.sdk.ValidationUtils.ensureValidCacheSet;
 import static momento.sdk.ValidationUtils.ensureValidKey;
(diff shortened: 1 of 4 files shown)


momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
```diff
@@ -678,6 +696,73 @@ public void onFailure(@Nonnull Throwable e) {
...
+          }
+        };
+
+    // Convert returned ListenableFuture to CompletableFuture
+    Futures.addCallback(
+        rspFuture,
+        new FutureCallback<_ListFetchResponse>() {
+          @Override
+          public void onSuccess(_ListFetchResponse rsp) {
+            if (rsp.getListCase().equals(_ListFetchResponse.ListCase.FOUND)) {
```
nand4011, line 728: Can you use rsp.hasFound() instead?

momento-sdk/src/main/java/momento/sdk/messages/CacheListFetchResponse.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Gets the retrieved values as a byte array.
+     *
+     * @return the values.
+     */
+    public List<byte[]> valueByteArrayList() {
+      List<byte[]> valuesAsByteArray = new ArrayList<>();
+      for (int i = 0; i < values.size(); i++) {
```
nand4011, line 36: Any reason you do it this way vs a stream?
`values.stream().map(ByteString::toByteArray).collect(Collectors.toList()))`
I think a stream can be slower in some cases.

There is also a question of whether we should perform this computation at constructor time vs doing it here vs memoizing it. @gautamomento @malandis any opinions? Guava has a Suppliers.memoize function and we are stuck with a guava dependency.

nand4011 (approved):

PR #189, opened 2023-03-28
feat: Add setAddElement and setFetch methods

nand4011:
Add setAddElement methods for string and byte[] elements.

Add setFetch.

Make seconds and milliseconds methods in CollectionTtl return Optionals because the internal ttl may be null.


rishtigupta (approved):
looks good to me!

PR #190, opened 2023-03-28
feat: Add setAddElements

nand4011:
Add setAddStringElements and setAddByteArrayElements methods. They have different names to avoid the method signatures being the same after type erasure.


rishtigupta (approved):
Could you rebase with main before pushing this commit? I just merged the listFetch API to main.

PR #194, opened 2023-03-29
feat: Add CacheListConcatenateFront API

rishtigupta:
### Commit 1
- Add CacheListConcatenateFront API
- Refactor CacheListConcatenateBack to match the front api
- Add/Refactor integ tests

### Commit 2
- Fix Integer to int for truncateFrontToSize & TruncateBackToSize
- As a result of above, had to split the String and ByteArray function names to avoid methods with same erasure exception
- Add tests to validate that same list can be added to front/back of the list

## Build

## Issue
#170

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/ListTest.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/ListTest.java
@@ -6,15 +6,20 @@ import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import momento.sdk.exceptions.InvalidArgumentException;
 import momento.sdk.messages.CacheListConcatenateBackResponse;
+import momento.sdk.messages.CacheListConcatenateFrontResponse;
 import momento.sdk.messages.CacheListFetchResponse;
 import momento.sdk.requests.CollectionTtl;
+import org.assertj.core.api.InstanceOfAssertFactories;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public class ListTest extends BaseTestClass {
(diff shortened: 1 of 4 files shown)


momento-sdk/src/intTest/java/momento/sdk/ListTest.java
```diff
@@ -231,4 +269,122 @@ public void shouldFetchAllValuesWhenListFetchWithInvalidIndices() {
 
     assertThat(cacheListFetchResponse).isInstanceOf(CacheListFetchResponse.Error.class);
   }
+
+  @Test
+  public void listConcatenateFrontStringHappyPath() {
+    List<String> oldValues = Arrays.asList("val1", "val2", "val3");
```
nand4011, line 275: You don't need to change this now, but we should make sure the same value can appear twice in a list.

momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -506,6 +507,48 @@ public CompletableFuture<CacheListConcatenateBackResponse> listConcatenateBack(
...
+   * @param truncateBackToSize If the list exceeds this length, remove excess from the front of the
+   *     list. Must be positive.
+   * @return Future containing the result of the list concatenate back operation.
+   */
+  public CompletableFuture<CacheListConcatenateFrontResponse> listConcatenateFront(
+      String cacheName,
+      String listName,
+      List<String> values,
+      CollectionTtl ttl,
+      Integer truncateBackToSize) {
```
nand4011, line 527: Why is this an Integer instead of an int?

rishtigupta: I didn't realize I was using Integer here. That was me experimenting before in the list concatenate back api. Let me correct both and push out another rev

nand4011 (approved):
Type erasure is a bummer sometimes. I like your method naming convention better than what I used, so I'll update the set methods to match it.

PR #196, opened 2023-03-29
chore: Remove open telemetry

nand4011:
Remove open telemetry. It could not be used as implemented because it was always set to null.

Remove unused methods from the stubs managers.


nand4011:
This is in response to an earlier misunderstanding on my part: [https://github.com/momentohq/client-sdk-java/pull/182#discussion_r1148031800](url)
I have already removed the tests but I didn't remove the otel functionality at the same time.

rishtigupta (approved):

PR #200, opened 2023-03-30
feat: Add listPushBack API

rishtigupta:
## PR Description
- Add ListPushBack API
- Add corresponding integ tests
- Remove error handling for ttl from functions since building the corresponding requests handle them.

## Build

## Issue
#170

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
@@ -386,12 +386,17 @@ final class ScsDataClient implements Closeable {
       String cacheName,
       String listName,
       List<String> values,
-      CollectionTtl ttl,
+      @Nullable CollectionTtl ttl,
       int truncateFrontToSize) {
     try {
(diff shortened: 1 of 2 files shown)


momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
```diff
@@ -388,9 +391,7 @@ CompletableFuture<CacheListConcatenateBackResponse> listConcatenateBackString(
     try {
       checkCacheNameValid(cacheName);
       checkListNameValid(listName);
-      if (ttl == null) {
```
nand4011, line 391: Where is this null check being done now?

rishtigupta: When building the requests, we use:
`.setTtlMilliseconds(ttl.toMilliseconds().orElse(itemDefaultTtl.toMillis()))`

which would mean to use the itemDefaultsTtl value if the tt.toMilliseconds is null, right?

nand4011 (approved):

PR #203, opened 2023-03-30
feat: Add method signatures that do not take ttl for list collections

rishtigupta:
## PR Description:
- Add method signatures that do not take ttl for list collections
- Refactor related integ tests

## Issue
#170, #201

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/CacheClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/CacheClient.java
@@ -522,6 +522,22 @@ public final class CacheClient implements Closeable {
    * @param cacheName Name of the cache to store the item in
    * @param listName The list in which the value is to be added.
    * @param values The elements to add to the list.
+   * @param truncateFrontToSize If the list exceeds this length, remove excess from the front of the
+   *     list. Must be positive.
+   * @return Future containing the result of the list concatenate back operation.
+   */
+  public CompletableFuture<CacheListConcatenateBackResponse> listConcatenateBackString(
+      String cacheName, String listName, List<String> values, int truncateFrontToSize) {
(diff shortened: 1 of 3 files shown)


momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
```diff
@@ -408,6 +408,25 @@ CompletableFuture<CacheListConcatenateBackResponse> listConcatenateBackString(
     }
   }
 
+  CompletableFuture<CacheListConcatenateBackResponse> listConcatenateBackString(
+      String cacheName, String listName, List<String> values, int truncateFrontToSize) {
+    try {
```
nand4011, line 413: Can these just call the other signature with null?

momento-sdk/src/main/java/momento/sdk/CacheClient.java
nand4011, line 1: I think we should also change the signatures with a ttl so that ttl is on the end to make them match. We could do that in a separate pass though.

rishtigupta: Yup, I think that is a good idea!

momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
rishtigupta: We can, I did this way because the set method creates different functions for handling null values and I wanted to be consistent with that. 
We can go either ways though. Do you want me to make the set methods without null to call set methods with null values?

rishtigupta: Got it, thanks. Will do.

nand4011 (approved):

PR #205, opened 2023-03-30
chore: Add new signatures without TTLs for set methods.

nand4011:


momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -473,10 +473,26 @@ public CompletableFuture<CacheIncrementResponse> increment(
...
+   * original elements of the set.
+   *
+   * @param cacheName Name of the cache to store the item in
+   * @param setName The set to add the element to.
+   * @param element The data to add to the set.
+   * @return Future containing the result of the add element operation.
+   */
+  public CompletableFuture<CacheSetAddElementResponse> setAddElement(
+      String cacheName, String setName, String element) {
+    return scsDataClient.setAddElement(cacheName, setName, element, null);
```
rishtigupta, line 493: I called the functions inside the cache client for these instead of passing the null ttls to sdsDataClient. Will approve this change and make a PR to be consistent with yours, since I think this way is better.

rishtigupta (approved):

PR #208, opened 2023-03-30
chore: Bring auth token handling in line with the .NET SDK

nand4011:
Create a CredentialProvider interface that supplies the auth token and the control and cache endpoint.

Make CredentialProvider implementations that read the auth token from a string or an environment variable.

Remove old auth token and endpoint handling logic.

Migrate tests to use CredentialProvider.


momento-sdk/src/intTest/java/momento/sdk/ListTest.java
```diff
@@ -27,16 +29,18 @@ public class ListTest extends BaseTestClass {
 
   private static final Duration FIVE_SECONDS = Duration.ofSeconds(5);
   private CacheClient target;
-  private String cacheName;
+
+  private final CredentialProvider credentialProvider =
+      new EnvVarCredentialProvider("TEST_AUTH_TOKEN");
+  private final String cacheName = System.getenv("TEST_CACHE_NAME");
```
rishtigupta, line 35: Why not use the EnvVarCredentialProvider for TEST_CACHE_NAME too?

momento-sdk/src/intTest/java/momento/sdk/SetTest.java
```diff
@@ -21,16 +23,18 @@
 
 public class SetTest {
   private static final Duration DEFAULT_TTL = Duration.ofSeconds(60);
-  private CacheClient client;
-  private String cacheName;
   private static final Duration FIVE_SECONDS = Duration.ofSeconds(5);
 
+  private final CredentialProvider credentialProvider =
+      new EnvVarCredentialProvider("TEST_AUTH_TOKEN");
+  private final String cacheName = System.getenv("TEST_CACHE_NAME");
```
rishtigupta, line 30: Same comment as above

momento-sdk/src/intTest/java/momento/sdk/BaseTestClass.java
```diff
@@ -20,8 +22,9 @@ static void beforeAll() {
   }
 
   private static void ensureTestCacheExists() {
+    final CredentialProvider credentialProvider = new EnvVarCredentialProvider("TEST_AUTH_TOKEN");
     try (CacheClient client =
-        CacheClient.builder(System.getenv("TEST_AUTH_TOKEN"), Duration.ofSeconds(10)).build()) {
+        CacheClient.builder(credentialProvider, Duration.ofSeconds(10)).build()) {
       client.createCache(System.getenv("TEST_CACHE_NAME"));
```
rishtigupta, line 28: Why not use EnvVarCredentialsProvider for getting TEST_CACHE_NAME too?

momento-sdk/src/intTest/java/momento/sdk/CacheDataPlaneClientSideTest.java
```diff
@@ -16,14 +18,15 @@
 final class CacheDataPlaneClientSideTest extends BaseTestClass {
 
   private static final Duration DEFAULT_ITEM_TTL_SECONDS = Duration.ofSeconds(60);
-  private String cacheName;
+
+  private final CredentialProvider credentialProvider =
+      new EnvVarCredentialProvider("TEST_AUTH_TOKEN");
+  private final String cacheName = System.getenv("TEST_CACHE_NAME");
```
rishtigupta, line 24: same comment as above

momento-sdk/src/intTest/java/momento/sdk/CacheDataPlaneTest.java
```diff
@@ -4,41 +4,77 @@
...
 /** Tests with Async APIs. */
 final class CacheDataPlaneTest extends BaseTestClass {
 
   private static final Duration DEFAULT_ITEM_TTL_SECONDS = Duration.ofSeconds(60);
-  private String authToken;
-  private String cacheName;
+
+  private final CredentialProvider credentialProvider =
+      new EnvVarCredentialProvider("TEST_AUTH_TOKEN");
+  private final String cacheName = System.getenv("TEST_CACHE_NAME");
```
rishtigupta, line 27: same comment

momento-sdk/src/intTest/java/momento/sdk/ListTest.java
nand4011: Cache name is only relevant for tests (and honestly I'm not sure why we even let the user specify it instead of picking a random one), so we don't need a way to get it in the main code.

rishtigupta: Ah, I see. 
Yeah, that makes sense if cache name is only being used in tests, we can specify a random name and is not required to be fetched from EnvVarCredentialsProvider. 

Thanks for clarifying!

rishtigupta (approved):

PR #211, opened 2023-03-31
feat: Add per-environment default configurations

nand4011:
Add a configuration object that contains settings for gRPC connections. It will eventually contain retry and middleware configs as well.

Add Laptop, InRegion, and LowLatency default configurations.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/CacheClientBuilder.java
+++ b/momento-sdk/src/main/java/momento/sdk/CacheClientBuilder.java
@@ -3,28 +3,40 @@ package momento.sdk;
 import java.time.Duration;
 import javax.annotation.Nonnull;
 import momento.sdk.auth.CredentialProvider;
+import momento.sdk.config.Configuration;
+import momento.sdk.config.transport.GrpcConfiguration;
+import momento.sdk.config.transport.TransportStrategy;
 
 /** Builder for {@link CacheClient} */
 public final class CacheClientBuilder {
 
   private final CredentialProvider credentialProvider;
+  private Configuration configuration;
   private final Duration itemDefaultTtl;
-  private Duration requestTimeout = null;
 
   CacheClientBuilder(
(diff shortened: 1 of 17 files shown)


momento-sdk/src/main/java/momento/sdk/CacheClientBuilder.java
```diff
@@ -3,28 +3,40 @@
...
 
   private final CredentialProvider credentialProvider;
+  private Configuration configuration;
   private final Duration itemDefaultTtl;
-  private Duration requestTimeout = null;
 
   CacheClientBuilder(
-      @Nonnull CredentialProvider credentialProvider, @Nonnull Duration itemDefaultTtl) {
+      @Nonnull CredentialProvider credentialProvider,
+      Configuration configuration,
```
rishtigupta, line 19: Should the configuration be @Nonnull too?

nand4011: Good catch.

momento-sdk/src/main/java/momento/sdk/config/Configurations.java
```diff
@@ -0,0 +1,87 @@
...
+     * Provides the latest recommended configuration for a dev environment.
+     *
+     * <p>This configuration may change in future releases to take advantage of improvements we
+     * identify for default configurations.
+     *
+     * @return the latest Laptop configuration
+     */
+    public static Configuration Latest() {
+      final TransportStrategy transportStrategy =
+          new StaticTransportStrategy(new GrpcConfiguration(Duration.ofMillis(15000)));
```
rishtigupta, line 31: Can we put the harcoded millisecond values as a global variable?
LATEST_LAPTOP_MILLIS, IN_REGION_MILLIS, LOW_LATENCY_MILLIS - something along these lines. Might help we we want to change these values in future?

rishtigupta: Got it, thanks for clarifying!

rishtigupta (approved):

PR #221, opened 2023-04-04
fix: remove string suffix, memoization, refactor java docs

rishtigupta:
## PR Description:
- remove string suffix
- remove memoization 
- refactor java docs
- rename items to elements to match the method signature in other sdks like .NET


cprice404 (approved):
looks good, thanks!!

PR #224, opened 2023-04-06
feat: Add didctionaryGetFields API

rishtigupta:
### Commit 1
- Add dictionaryGetFields API
- Add field parameter to CacheDictionaryGetFieldResponse.Error constructor (required for implementing the plural case)
- Add integ tests

### Commit 2
- Remove elements and fields internal representation from response class
- Use the responseList to get the map for desired data types
- fix minor bugs/ addressed comments on commit 1

## Issue
#171

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
@@ -923,14 +923,14 @@ final class ScsDataClient implements Closeable {
 
   private ByteString convert(String stringToEncode) {
     if (stringToEncode == null) {
-      return ByteString.copyFromUtf8("null");
+      return ByteString.EMPTY;
     }
     return ByteString.copyFromUtf8(stringToEncode);
   }
 
   private ByteString convert(byte[] bytes) {
     if (bytes == null) {
-      return ByteString.copyFromUtf8("null");
+      return ByteString.EMPTY;
     }
     return ByteString.copyFrom(bytes);
   }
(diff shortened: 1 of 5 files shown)


momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
```diff
@@ -879,15 +881,57 @@ CompletableFuture<CacheDictionaryGetFieldResponse> dictionaryGetField(
...
+      return sendDictionaryGetFields(
+          cacheName, convert(dictionaryName), convertByteArrayList(fields));
+    } catch (Exception e) {
+      return CompletableFuture.completedFuture(
+          new CacheDictionaryGetFieldsResponse.Error(CacheServiceExceptionMapper.convert(e)));
     }
   }
 
   private ByteString convert(String stringToEncode) {
+    if (stringToEncode == null) {
```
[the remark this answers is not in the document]
rishtigupta: Ah, good call. The other sdks return ByteString.EMPTY if the string/byte to convert is null

momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryGetFieldsResponse.java
```diff
@@ -0,0 +1,174 @@
...
+        counter++;
+      }
+    }
+
+    /**
+     * Gets the retrieved dictionary of string keys and string values.
+     *
+     * @return the dictionary.
+     */
+    public Map<String, String> valueDictionaryStringString() {
```
cprice404, line 55: can we add an alias for this one that is just named `valueDictionary`?

cprice404 (approved):
lgtm!  would love to have an alias for the 90% value accessor but not a blocker

momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryGetFieldsResponse.java
rishtigupta: Will handle this in a separate PR for all the apis that use this type of accessors.

PR #232, opened 2023-04-07
chore: Add alias for string accessors in response classes

rishtigupta:
## PR:
Add alias for string accessors in response classes

## Issue
#230

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryFetchResponse.java
+++ b/momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryFetchResponse.java
@@ -55,6 +55,19 @@ public interface CacheDictionaryFetchResponse {
     }
 
     /**
+     * Gets the retrieved values as a dictionary of UTF-8 string keys and values.
+     *
+     * @return the dictionary.
+     */
+    public Map<String, String> valueDictionary() {
+      return byteStringKeysValues.entrySet().stream()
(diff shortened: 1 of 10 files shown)


momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryFetchResponse.java
```diff
@@ -54,6 +54,19 @@ public Map<String, String> valueDictionaryStringString() {
                   entry -> entry.getValue().toStringUtf8()));
     }
 
+    /**
+     * Gets the retrieved values as a dictionary of UTF-8 string keys and values.
+     *
+     * @return the dictionary.
+     */
+    public Map<String, String> valueDictionary() {
+      return byteStringKeysValues.entrySet().stream()
```
nand4011, line 63: Can this just call `valueDictionaryStringString()`?

nand4011 (commented):
If we're going to keep both implementations I would make these call the original method instead of reimplementing the logic.

momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryFetchResponse.java
rishtigupta: That function exists. This is just an alias of that function

nand4011 (approved):

PR #233, opened 2023-04-07
feat: Add sorted set put and fetch methods

nand4011:
Add sortedSetPutElement and sortedSetPutElements.

Add sortedSetFetchByRank and sortedSetFetchByScore.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/CacheClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/CacheClient.java
@@ -49,6 +49,7 @@ import momento.sdk.messages.FlushCacheResponse;
 import momento.sdk.messages.ListCachesResponse;
 import momento.sdk.messages.ListSigningKeysResponse;
 import momento.sdk.messages.RevokeSigningKeyResponse;
+import momento.sdk.messages.SortOrder;
 import momento.sdk.requests.CollectionTtl;
 
 /** Client to perform operations against the Momento Cache Service */
@@ -828,14 +829,16 @@ public final class CacheClient implements Closeable {
    * @param endRank - The rank of the last element to fetch. This rank is exclusive, ie the element
    *     at this rank will not be fetched. Defaults to null, which fetches up until and including
    *     the last element.
+   * @param order - The order to fetch the elements in. Defaults to ascending.
    * @return Future containing the result of the fetch operation.
    */
   public CompletableFuture<CacheSortedSetFetchResponse> sortedSetFetchByRank(
       String cacheName,
       String sortedSetName,
       @Nullable Integer startRank,
-      @Nullable Integer endRank) {
-    return scsDataClient.sortedSetFetchByRank(cacheName, sortedSetName, startRank, endRank);
+      @Nullable Integer endRank,
+      @Nullable SortOrder order) {
+    return scsDataClient.sortedSetFetchByRank(cacheName, sortedSetName, startRank, endRank, order);
   }
 
   /**
@@ -847,7 +850,7 @@ public final class CacheClient implements Closeable {
    */
   public CompletableFuture<CacheSortedSetFetchResponse> sortedSetFetchByRank(
       String cacheName, String sortedSetName) {
-    return scsDataClient.sortedSetFetchByRank(cacheName, sortedSetName, null, null);
+    return scsDataClient.sortedSetFetchByRank(cacheName, sortedSetName, null, null, null);
   }
 
   /**
@@ -859,6 +862,7 @@ public final class CacheClient implements Closeable {
    *     infinity.
    * @param maxScore - The maximum score (inclusive) of the elements to fetch. Defaults to positive
    *     infinity.
+   * @param order - The order to fetch the elements in. Defaults to ascending.
    * @param offset - The number of elements to skip before returning the first element. Defaults to
    *     0. Note: this is not the rank of the first element to return, but the number of elements of
    *     the result set to skip before returning the first element.
@@ -870,10 +874,11 @@ public final class CacheClient implements Closeable {
       String sortedSetName,
       @Nullable Double minScore,
       @Nullable Double maxScore,
+      @Nullable SortOrder order,
       @Nullable Integer offset,
       @Nullable Integer count) {
     return scsDataClient.sortedSetFetchByScore(
-        cacheName, sortedSetName, minScore, maxScore, offset, count);
+        cacheName, sortedSetName, minScore, maxScore, order, offset, count);
   }
 
   /**
@@ -889,7 +894,8 @@ public final class CacheClient implements Closeable {
    */
   public CompletableFuture<CacheSortedSetFetchResponse> sortedSetFetchByScore(
       String cacheName, String sortedSetName, @Nullable Integer offset, @Nullable Integer count) {
-    return scsDataClient.sortedSetFetchByScore(cacheName, sortedSetName, null, null, offset, count);
+    return scsDataClient.sortedSetFetchByScore(
+        cacheName, sortedSetName, null, null, null, offset, count);
   }
 
   /**
--- a/momento-sdk/src/main/java/momento/sdk/CacheClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/CacheClient.java
@@ -39,6 +39,9 @@ import momento.sdk.messages.CacheSetIfNotExistsResponse;
 import momento.sdk.messages.CacheSetRemoveElementResponse;
 import momento.sdk.messages.CacheSetRemoveElementsResponse;
 import momento.sdk.messages.CacheSetResponse;
+import momento.sdk.messages.CacheSortedSetFetchResponse;
+import momento.sdk.messages.CacheSortedSetPutElementResponse;
+import momento.sdk.messages.CacheSortedSetPutElementsResponse;
 import momento.sdk.messages.CreateCacheResponse;
 import momento.sdk.messages.CreateSigningKeyResponse;
 import momento.sdk.messages.DeleteCacheResponse;
@@ -46,6 +49,7 @@ import momento.sdk.messages.FlushCacheResponse;
 import momento.sdk.messages.ListCachesResponse;
 import momento.sdk.messages.ListSigningKeysResponse;
 import momento.sdk.messages.RevokeSigningKeyResponse;
+import momento.sdk.messages.SortOrder;
 import momento.sdk.requests.CollectionTtl;
 
 /** Client to perform operations against the Momento Cache Service */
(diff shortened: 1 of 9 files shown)


momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -677,6 +681,219 @@ public CompletableFuture<CacheSetFetchResponse> setFetch(String cacheName, Strin
...
+  }
+
+  /**
+   * Fetch all elements in the given sorted set by index (rank) in ascending order.
+   *
+   * @param cacheName - The cache containing the sorted set.
+   * @param sortedSetName - The sorted set to fetch from.
+   * @return Future containing the result of the fetch operation.
+   */
+  public CompletableFuture<CacheSortedSetFetchResponse> sortedSetFetchByRank(
```
nand4011, line 851: I added this and the equivalent sortedSetFetchByScore because I assume people will want to fetch the entire set regularly.

momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -677,6 +681,219 @@ public CompletableFuture<CacheSetFetchResponse> setFetch(String cacheName, Strin
...
+  }
+
+  /**
+   * Fetch all elements in the given sorted set by score in ascending order.
+   *
+   * @param cacheName - The cache containing the sorted set.
+   * @param sortedSetName - The sorted set to fetch from.
+   * @return Future containing the result of the fetch operation.
+   */
+  public CompletableFuture<CacheSortedSetFetchResponse> sortedSetFetchByScore(
```
nand4011, line 891: Should this take an offset and count since the other pared down fetch method already fetches the whole set?

momento-sdk/src/main/java/momento/sdk/messages/CacheSortedSetFetchResponse.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Constructs a sorted set fetch hit with a list of encoded elements.
+     *
+     * @param elements the retrieved elements.
+     */
+    public Hit(List<_SortedSetElement> elements) {
+      this.elements = elements;
+    }
+
+    public NavigableSet<ScoredElement<String>> elementsSet() {
```
nand4011, line 28: I added these navigable set and ordered list methods to compare with the ordered map method.

```java
{
  // The map won't update its order if the scores are changed
  final Map<String, Double> scoredMap = hit.elementsMap();
  // iterate through in order
  for (Map.Entry<String, Double> entry : scoredMap.entrySet()) {
    final String element = entry.getKey();
    final double score = entry.getValue();
    // ...
  }

  // get score (doesn't work for byte[])
  final Double score = scoredMap.get("thing I want");
}

{
  // The list won't update its order if the scores are changed
  final List<ScoredElement<String>> scoredList = hit.elementsList();
  // iterate through in order
  for (ScoredElement<String> scoredElement : scoredList) {
    final String element = scoredElement.getElement();
    final double score = scoredElement.getScore();
    // ...
  }

  // get score
  final Optional<Double> score =
      scoredList.stream()
          .filter(se -> se.getElement().equals("thing I want"))
          .map(ScoredElement::getScore)
          .findFirst();
}

{
  // Will maintain order when modifying set. Doesn't make sense to specify
  // ascending/descending in the method call because this can be traversed either way.
  final NavigableSet<ScoredElement<String>> scoredSet = hit.elementsSet();
  // iterate through in ascending order
  for (ScoredElement<String> scoredElement : scoredSet) {
    final String element = scoredElement.getElement();
    final double score = scoredElement.getScore();
    // ...
  }

  // iterate through in descending order
  for (ScoredElement<String> scoredElement : scoredSet.descendingSet()) {
    final String element = scoredElement.getElement();
    final double score = scoredElement.getScore();
    // ...
  }

  // get score
  final Optional<Double> score =
      scoredSet.stream()
          .filter(se -> se.getElement().equals("thing I want"))
          .map(ScoredElement::getScore)
          .findFirst();
}
```

nand4011: @cprice404 would you like a more in-depth example?

momento-sdk/src/main/java/momento/sdk/CacheClient.java
cprice404: 👍

momento-sdk/src/main/java/momento/sdk/messages/CacheSortedSetFetchResponse.java
cprice404: no this is great, thanks.

One thing to call out here is that I don't know if these `stream` parts are that relevant; if the user wants a score for an individual element they would probably just call one of the other methods like `GetScore`/`GetScores`.  If they want to do something will all the scores then they would just access the score of the current element inside of a loop or `map`.  So personally I wouldn't weigh that stanza very heavily in evaluating the ergonomics of each of these examples.

Given that, I really like your `NavigableSet` solution.  That has my vote.

cprice404 (commented):
Looking good!  I left my thoughts on the return data type, and a question about Optional.  Otherwise lgtm

momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -677,6 +681,219 @@ public CompletableFuture<CacheSetFetchResponse> setFetch(String cacheName, Strin
...
+   *     at this rank will not be fetched. Defaults to null, which fetches up until and including
+   *     the last element.
+   * @param order - The order to fetch the elements in. Defaults to ascending.
+   * @return Future containing the result of the fetch operation.
+   */
+  public CompletableFuture<CacheSortedSetFetchResponse> sortedSetFetchByRank(
+      String cacheName,
+      String sortedSetName,
+      @Nullable Integer startRank,
+      @Nullable Integer endRank,
```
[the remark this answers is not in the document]
nand4011: Optionals in Java as method parameters generally feel off to me. They don't add null safety and they require an extra layer of wrapping to use. I know they were explicitly not idiomatic when optional was released. My instincts might be wrong though.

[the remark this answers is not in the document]
cprice404: okay.  I don't consider myself up-to-date on java best practices on this, and a quick google search yielded articles like [this one](https://medium.com/javarevisited/4-reasons-why-you-should-use-java-optional-or-not-4e44d51a9645) which agree with your take.  So I am happy to go with your instinct here.  It might be a thing to call out explicitly in your bug bash doc though to see if others with Java backgrounds want to weigh in on it.

momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -829,16 +828,14 @@ public CompletableFuture<CacheSortedSetPutElementsResponse> sortedSetPutElements
...
    * @return Future containing the result of the fetch operation.
    */
   public CompletableFuture<CacheSortedSetFetchResponse> sortedSetFetchByRank(
       String cacheName,
       String sortedSetName,
       @Nullable Integer startRank,
-      @Nullable Integer endRank,
-      @Nullable SortOrder order) {
-    return scsDataClient.sortedSetFetchByRank(cacheName, sortedSetName, startRank, endRank, order);
+      @Nullable Integer endRank) {
```
cprice404, line 837: removing the SortOrder seems wrong to me.  Don't we have to tell the server the sort order so that the indices we are asking for will be correct?  Or is it applying the indices first and then applying the sort order afterward (in which case what you are doing here is probably okay)?

If you know the answer to this can you please update the notes here in the spec:

https://www.notion.so/momentohq/SDK-API-Methods-Spec-cf12e4db2535405ca2acc3608814b083#238e5fc189a5498392e9dc2d0fd45e8b

and maybe also check the public dev docs to make sure they are clear about this?

momento-sdk/src/main/java/momento/sdk/CacheClient.java
cprice404: that's probably okay and we could add another signature without those in it in the future if we wanted, right?

momento-sdk/src/main/java/momento/sdk/messages/CacheSortedSetFetchResponse.java
```diff
@@ -25,54 +23,17 @@ public Hit(List<_SortedSetElement> elements) {
...
-     * determined by the options used in the method call.
+     * Gets a navigable set of the retrieved elements and their scores. The set is in ascending
+     * order by default and can be viewed in descending order with {@link
+     * NavigableSet#descendingSet()}.
      *
-     * @return An ordered map of byte[] elements to scores
+     * @return An ordered set of elements to scores
      */
-    public Map<byte[], Double> elementsByteArrayMap() {
+    public NavigableSet<ScoredElement> elementsSet() {
```
cprice404, line 35: still curious to know if someone could pass this back in as input.  I'm not reviewing this PR as thoroughly as I normally would because it's late and we're pre-1.0 and I don't want to keep you waiting, so not a blocker, but food for thought :)

cprice404 (commented):
i have a clarifying question about removing the SortOrder param but I'm +1 pending the answer to that question

momento-sdk/src/main/java/momento/sdk/CacheClient.java
nand4011: We could indeed.

momento-sdk/src/main/java/momento/sdk/CacheClient.java
nand4011: I hadn't considered the interaction between sort order and the indices. If that's the case then I think we need to use a list instead of a NavigableSet since the set will have the same default sort order regardless of what order the user wants.

momento-sdk/src/main/java/momento/sdk/messages/CacheSortedSetFetchResponse.java
nand4011: Right now you could not. I can update ScoredElement to be more usable by consumers. This opens a more broad question about encapsulating classes as a solution to the type erasure problem, since ScoredElement effectively is such a wrapping class.

cprice404: let's keep an eye out for this in bug bash.  I think we should assign owners to different collection types to try to get coverage.  you should remind me to choose sortedset (possibly along with some other folks if we have enough volunteers).

cprice404 (approved):

PR #235, opened 2023-04-11
feat: Add sortedSetGetRank

nand4011:
Add sortedSetGetRank. This returns the rank of the given sorted set element.

Add generic code for executing a grpc function. This should cut down on boilerplate in the internal data client.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
@@ -571,13 +571,13 @@ final class ScsDataClient implements Closeable {
   }
 
   CompletableFuture<CacheSortedSetGetRankResponse> sortedSetGetRank(
-      String cacheName, String sortedSetName, String element) {
+      String cacheName, String sortedSetName, String element, @Nullable SortOrder order) {
     try {
       checkCacheNameValid(cacheName);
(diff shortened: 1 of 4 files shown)


momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
```diff
@@ -1890,6 +1923,32 @@ public void onFailure(@Nonnull Throwable e) {
     return returnFuture;
   }
 
+  private CompletableFuture<CacheSortedSetGetRankResponse> sendSortedSetGetRank(
```
nand4011, line 1926: I figured we can migrate to something like this to cut down on the repetition in the internal client. I could do it all on one line but it is a bit confusing without named arguments.
If there's more I can do to clean this up let me know. I'm sure there's a more object oriented way to encapsulate this that will cut down our 3,000 line client even more.

cprice404: the main thing I have been thinking of for breaking up these clients in the future is pulling the code for each data type out into a separate file; either using abstract classes or decorator pattern or something.  I think that will help a fair amount when we decide it's worth it.

cprice404 (dismissed):
🙈

cprice404 (approved):

PR #257, opened 2023-04-17
chore: add fromString and fromEnvVar methods to CredentialProvider

nand4011:
Add fromString and fromEnvVar methods to CredentialProvider.

Change most instances of CredentialProvider in tests to use the new methods.

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/BaseTestClass.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/BaseTestClass.java
@@ -2,7 +2,6 @@ package momento.sdk;
 
 import java.time.Duration;
 import momento.sdk.auth.CredentialProvider;
-import momento.sdk.auth.EnvVarCredentialProvider;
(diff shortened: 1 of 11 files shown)


momento-sdk/src/intTest/java/momento/sdk/BaseTestClass.java
```diff
@@ -23,7 +22,7 @@ static void beforeAll() {
   }
 
   private static void ensureTestCacheExists() {
-    final CredentialProvider credentialProvider = new EnvVarCredentialProvider("TEST_AUTH_TOKEN");
+    final CredentialProvider credentialProvider = CredentialProvider.fromEnvVar("TEST_AUTH_TOKEN");
```
cprice404, line 25: In the other SDKs all of the cred provider constructors provide a way to override the control and data plane endpoints, including these factory functions.  I do think we will need that here but it is a backward compatible addition, so not a blocker for this PR.

cprice404 (approved):

PR #259, opened 2023-04-17
chore: Create v1 client configurations.

nand4011:
Create v1 versions of the laptop, in-region, and low-latency configurations. These are guaranteed not to change in the future after the 1.0 SDK is released.


momento-sdk/src/main/java/momento/sdk/config/Configurations.java
```diff
@@ -79,6 +99,16 @@ private LowLatency(TransportStrategy transportStrategy) {
...
+      return LowLatency.V1();
+    }
+
+    /**
+     * Provides v1 configuration for a low-latency environment. This configuration is guaranteed not
+     * to change in future releases of the Momento Java SDK.
+     *
+     * @return the v1 low-latency configuration
+     */
+    public static Configuration V1() {
```
cprice404, line 111: nit for all of these - I think the `PascalCasing` was for .NET.  It might be more idiomatic to use `camelCasing` for these in Java, but not a big deal

nand4011: Good catch.

PR #260, opened 2023-04-18
chore: Remove ByteBuffer methods and fix warnings

nand4011:
Remove client methods that take a ByteBuffer. They can be added later if desired.

Fix warnings in ScsDataClient.


cprice404 (approved):
so long

Issue #248, opened 2023-04-17
Finalize SortedSet collection type


nand4011:
- Audit the set methods against the .NET SDK.
- Ensure consistent naming for fields, values, elements, etc. is used.
- Ensure that methods that take a collection take the most general collection that makes sense.
- Ensure that methods that take a collection behave in an expected way when given an empty collection.
- Only @Nullable annotate arguments if there isn't a signature that leaves it out. If there isn't, check if there should be.

Specific fixes:
- Make a fetchByScore method that fetches the whole set.
- Make the fetchByScore method that takes offset and count take sortOrder as well.
- Make sortOrder the first argument after setName in the sortBy* methods.
- Make fetchByScore take offset and count before minScore and maxScore.
- Make constructors for ScoredElement that take a String and byte[].
- Make a putElements signature that takes a collection of ScoredElements.
- Check if getScores returns ordered results. If not, make its return type a Set.

malandis:
✅ API names: https://github.com/momentohq/client-sdk-java/pull/262

malandis:
🟨  Do the methods take the most general collection that makes sense?
1. ✅ Put: `Map<., Double>`. The alternative would be an iterable of `Map.Entry<., Double>`. This is not consistent with the java `Map` interface, and would likely feel unfamiliar to a user.
2. 🟨 GetScores: `Set<.>`. An alternative would be `Iterable<.>`. I prefer this since `Set` imposes an additional constraint that a user may not care about. Ie they could have an `Array`, which may or may not contain duplicates (and may not matter); having to form a `Set` from the `Array` is unnecessary extra work.
3. 🟨 RemoveElements: `Set<.>`. Same as above. Prefer `Iterable<.>`

malandis:
✅ collection inputs: https://github.com/momentohq/client-sdk-java/pull/263

malandis:
✅ empty collection behavior: in general we do not penalize the user if they pass an empty collection

malandis:
✅ sorted set fetch by score: get the whole sorted set. https://github.com/momentohq/client-sdk-java/pull/267

malandis:
✅  Check if getScores returns ordered results.

PR #262, opened 2023-04-18
chore: normalize sorted set api names

malandis:
Many sorted set APIs alter an element, as indexed by a
`value`. Consistent with other SDKs, we call the argument `value`, not
`element`, since `element` can be confused with the value+score
combination.

work towards https://github.com/momentohq/client-sdk-java/issues/248


cprice404 (approved):
🙈

PR #263, opened 2023-04-19
chore: audit sorted set collection arguments

malandis:
Lift `sortedSetGetScores` and `sortedSetRemoveElements` argument `values` from `Set` to `Iterable`. Generally speaking the service does not impose uniqueness constraints on collection inputs, so a narrow `Set` input is not necessary. A user may pass in any iterable container, which may or may not contain duplicates.

Work towards: https://github.com/momentohq/client-sdk-java/issues/248

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
@@ -91,6 +91,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import momento.sdk.auth.CredentialProvider;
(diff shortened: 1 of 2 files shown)


momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
```diff
@@ -619,28 +620,29 @@ CompletableFuture<CacheSortedSetGetScoreResponse> sortedSetGetScore(
   }
 
   CompletableFuture<CacheSortedSetGetScoresResponse> sortedSetGetScores(
-      String cacheName, String sortedSetName, Set<String> elements) {
+      String cacheName, String sortedSetName, Iterable<String> values) {
     try {
       checkCacheNameValid(cacheName);
       checkSetNameValid(sortedSetName);
-      ensureValidValue(elements);
+      ensureValidValue(values);
```
bruuuuuuuce, line 627: I always get tripped up on this. The `value` is the name of the element stored in the sorted set?

bruuuuuuuce (approved):
:shipit:

nand4011 (approved):

momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
malandis: Element refers to the value and score. An element can be indexed by its value or rank.

PR #264, opened 2023-04-19
chore: audit sorted set internal arguments

malandis:
This PR corrects some of the sorted set names used internally.


bruuuuuuuce (approved):
:shipit:

PR #265, opened 2023-04-19
fix: sorted set name validation message should mention sorted sets

malandis:
The previous validation message for checking the sorted set name was
the message for sets. This fix adds a specific message for sorted sets
to differentiate from the set collection.


bruuuuuuuce (approved):
:shipit:

PR #268, opened 2023-04-19
chore: Update list method documentation.

nand4011:
Update the Javadocs for the list methods.

Add Nullable and Nonnull annotations to all list method arguments.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/CacheClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/CacheClient.java
@@ -1115,393 +1115,453 @@ public final class CacheClient implements Closeable {
   }
 
   /**
-   * Concatenates values to the back of the list.
+   * Adds the given values to the back of a list.
    *
-   * @param cacheName Name of the cache to store the item in
-   * @param listName The list in which the value is to be added.
-   * @param values The elements to add to the list.
+   * @param cacheName The cache containing the list.
+   * @param listName The list in which to add the values.
+   * @param values The values to add to the list.
    * @param truncateFrontToSize If the list exceeds this length, remove excess from the front of the
-   *     list. Must be positive.
-   * @param ttl Time to Live for the item in Cache. This ttl takes precedence over the TTL used when
-   *     building a cache client {@link CacheClient#builder(CredentialProvider, Configuration,
-   *     Duration)}
-   * @return Future containing the result of the list concatenate back operation.
+   *     list. Must be positive. Will not truncate if not provided.
+   * @param ttl TTL for the set in cache. This TTL takes precedence over the TTL used when
+   *     initializing a cache client. Defaults to the client's TTL if not provided.
+   * @return Future containing the result of the list concatenate back operation: {@link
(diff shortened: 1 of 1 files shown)


momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -1115,393 +1115,458 @@ public CompletableFuture<CacheSortedSetRemoveElementsResponse> sortedSetRemoveEl
...
+   *     CacheListConcatenateBackResponse.Error}.
    */
   public CompletableFuture<CacheListConcatenateBackResponse> listConcatenateBack(
-      String cacheName,
-      String listName,
-      List<String> values,
-      int truncateFrontToSize,
+      @Nonnull String cacheName,
+      @Nonnull String listName,
+      @Nonnull Iterable<String> elements,
```
nand4011, line 1134: Do we want to use elements or values for sets/lists? It's easy for dictionaries since a value is a piece of an element but here they seem synonymous.

momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -1115,393 +1115,458 @@ public CompletableFuture<CacheSortedSetRemoveElementsResponse> sortedSetRemoveEl
...
+   * @return Future containing the result of the list concatenate back operation: {@link
+   *     CacheListConcatenateBackResponse.Success} or {@link
+   *     CacheListConcatenateBackResponse.Error}.
    */
   public CompletableFuture<CacheListConcatenateBackResponse> listConcatenateBack(
-      String cacheName,
-      String listName,
-      List<String> values,
-      int truncateFrontToSize,
+      @Nonnull String cacheName,
```
nand4011, line 1132: @cprice404 What do you think of the way these look with fully annotated arguments? Nonnull is helpful for generating good warnings for the user and it would look strange to not annotate the Nullable arguments.

malandis (commented):
I prefer the annotation coverage on the public API.

re the names: we use `value` and `values` in every other SDK

momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -1115,393 +1115,453 @@ public CompletableFuture<CacheSortedSetRemoveElementsResponse> sortedSetRemoveEl
...
    */
   public CompletableFuture<CacheListRemoveValueResponse> listRemoveValue(
-      String cacheName, String listName, String value) {
+      @Nonnull String cacheName, @Nonnull String listName, @Nonnull String value) {
     return scsDataClient.listRemoveValue(cacheName, listName, value);
   }
 
   /**
-   * Removes value from the given list.
+   * Removes an value from a list.
```
malandis, line 1537:
```suggestion
   * Removes a value from a list.
```

nand4011: Missed a few of those when changing element to value.

malandis (approved):
✨

PR #270, opened 2023-04-20
chore: pare down `sortedSetFetchByScore` overloads

malandis:
Until a future ask comes in, we will suppose the following overloads
for `sortedSetFetchByScore`:

- (nothing) fetch the whole sorted set
- (common options) minScore, maxScore, sortOrtder
- (all options) minScore, maxScore, sortOrder, limit, offset

work towards: https://github.com/momentohq/client-sdk-java/issues/248

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/SortedSetTest.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/SortedSetTest.java
@@ -435,8 +435,7 @@ public class SortedSetTest extends BaseTestClass {
 
     // Full set ascending, end index larger than set
     assertThat(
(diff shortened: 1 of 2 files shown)


cprice404 (approved):
wfm if Nate's good with it, not sure why the tests are failing

nand4011 (approved):

nand4011:
The intermittent failures started after I switched the control methods to async. There may be a place where createCache is called that I missed adding join() to. Making them async didn't change the syntax of the method so any place that was calling one of them without doing something with the response would still compile but not behave as expected.

PR #271, opened 2023-04-20
feat: add `FetchByScoreOptions` builder

malandis:
To allow more flexible invocations of `sortedSetFetchByScore`, we
introduce an options class and builder class. That way a user can
selectively set any subset of the options.

Note: we can merge this upon a customer request.


cprice404:
thanks for posting this!  we don't need to keep it as an open PR, we can just keep the branch around.  Would also be good to push the branch to your fork so we have another copy of it.  After that you can close this PR.  Thanks!

PR #277, opened 2023-04-20
chore: Update the dictionary javadoc

nand4011:
Update the javadoc for the dictionary methods.

Update the javadoc for the dictionary responses.

Change the field/value/element language in the dictionary response getters.

Annotate the dictionary user-facing and internal methods with Nullable and Nonnull.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryFetchResponse.java
+++ b/momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryFetchResponse.java
@@ -15,20 +15,20 @@ public interface CacheDictionaryFetchResponse {
     private final Map<ByteString, ByteString> fieldsToValues;
 
     /**
-     * Constructs a dictionary fetch hit with a list of encoded keys and values.
+     * Constructs a dictionary fetch hit with a map of encoded fields to values.
      *
-     * @param fieldsToValues the retrieved dictionary.
+     * @param fieldsToValues The retrieved map.
      */
     public Hit(Map<ByteString, ByteString> fieldsToValues) {
       this.fieldsToValues = fieldsToValues;
     }
 
     /**
-     * Gets the retrieved values as a dictionary of UTF-8 string keys and values.
+     * Gets the retrieved elements as a map of UTF-8 string fields to UTF-8 string values.
      *
-     * @return the dictionary.
+     * @return The map.
      */
-    public Map<String, String> valueDictionaryStringString() {
+    public Map<String, String> valueMapStringString() {
       return fieldsToValues.entrySet().stream()
           .collect(
               Collectors.toMap(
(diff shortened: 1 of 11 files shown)


momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryFetchResponse.java
```diff
@@ -15,20 +15,20 @@ class Hit implements CacheDictionaryFetchResponse {
...
 
     /**
-     * Gets the retrieved values as a dictionary of UTF-8 string keys and values.
+     * Gets the retrieved elements as a map of UTF-8 string fields to UTF-8 string values.
      *
-     * @return the dictionary.
+     * @return The map.
      */
-    public Map<String, String> valueDictionaryStringString() {
+    public Map<String, String> elementsStringString() {
```
malandis, line 31: Do we not prefix the payload we `value`?

nand4011: I'm not sure I understand. I changed this to elements because the map is a combination of fields and values. I think that 'dictionary' was in there because .NET has a Dictionary type.

cprice404: sorry @nand4011 this is my bad, I'm sure the way we talked about argument names yesterday was misleading and I left this part out.

Everything we talked about yesterday I think still holds for argument names, but for all of our Response.Hit objects, we have adopted the convention of always using the `.value` prefix for all data accessors.  Across all of the data types.  The intent is that any time a user is trying to get data out of a Response.Hit they will know that they can just type `.val` and then the autocomplete will show them all of the choices.

momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryGetFieldsResponse.java
```diff
@@ -10,35 +10,36 @@
...
 
     /**
-     * Gets the retrieved dictionary of string keys and string values.
+     * Gets the retrieved map of string fields and string values.
      *
-     * @return the dictionary.
+     * @return The map.
      */
-    public Map<String, String> valueDictionaryStringString() {
+    public Map<String, String> elementsStringString() {
```
cprice404, line 42: comment above applies to this response class as well.

cprice404 (commented):
lgtm other than flipping the two Response.Hit accessor names back to `.value*`.

momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryFetchResponse.java
nand4011: Thanks for clarifying.

momento-sdk/src/main/java/momento/sdk/messages/CacheDictionaryFetchResponse.java
```diff
@@ -37,20 +37,20 @@ public Map<String, String> valueDictionaryStringString() {
...
 
     /**
-     * Gets the retrieved value as a dictionary of UTF-8 String keys and byte array values
+     * Gets the retrieved elements as a map of UTF-8 string fields to byte array values.
      *
-     * @return the dictionary.
+     * @return The map.
      */
-    public Map<String, byte[]> valueDictionaryStringBytes() {
+    public Map<String, byte[]> valueMapStringBytes() {
```
malandis, line 53: "Bytes" or "ByteArray"? I am looking for consistency across `Set` and `SortedSet`. I believe we want `ByteArray` but let me know if a decision was made for these compound names.

nand4011: ByteArray makes more sense; I'll change it. I imagine Bytes was used because valueDictionaryByteArrayByteArray is unpleasantly long.

malandis (approved):

cprice404 (approved):

PR #282, opened 2023-04-24
chore: Rename responses to remove 'Cache' and make them consistent

nand4011:
Rename all responses to remove 'Cache' from the front, except for control responses that operate on caches.

Move responses into a cache package and into packages for each collection.


nand4011:
IntelliJ refactoring is a lifesaver.

cprice404 (approved):
🙈

momento-sdk/src/intTest/java/momento/sdk/CacheControlPlaneTest.java
```diff
@@ -12,12 +12,12 @@
...
-import momento.sdk.responses.DeleteCacheResponse;
-import momento.sdk.responses.ListCachesResponse;
-import momento.sdk.responses.ListSigningKeysResponse;
-import momento.sdk.responses.RevokeSigningKeyResponse;
+import momento.sdk.responses.cache.control.CacheCreateResponse;
+import momento.sdk.responses.cache.control.CacheDeleteResponse;
+import momento.sdk.responses.cache.control.CacheListResponse;
+import momento.sdk.responses.cache.signing.SigningKeyCreateResponse;
+import momento.sdk.responses.cache.signing.SigningKeyListResponse;
+import momento.sdk.responses.cache.signing.SigningKeyRevokeResponse;
```
malandis, line 20: I previously commented that these are control plane responses, so with the previous convention would be grouped there with `Cache` as a prefix, but makes sense to group in their own package too.

malandis (approved):

PR #286, opened 2023-04-26
chore: Update examples for the 1.0 SDK release

nand4011:
Make the examples compatible with the 1.0 Java SDK.

Move client creation out of a parent class and into each example.

Log the error code when an error occurs.

Add logging to the examples other than the basic one.


README.md
```diff
@@ -67,75 +67,102 @@ Add our dependency your `pom.xml`
 
 Checkout our [examples](./examples/README.md) directory for complete examples of how to use the SDK.
 
-Here is a quickstart you can use in your own project:
+Here is a basic example you can use to get started:
```
cprice404, line 70: nit: if we are using the README template engine for this repo then I wouldn't expect to see changes to this file in a PR.  They will be overwritten when the template engine runs.  Not a problem if the changes are identical to the ones in the template but just a thing to be aware of.

README.template.md
```diff
@@ -53,7 +53,7 @@ Add our dependency your `pom.xml`
```
cprice404, line 53: kinda does look like there are a bunch of changes to the README that aren't in this template though?

cprice404 (commented):
looks good other than the question about the readme template

malandis (commented):
Looks good. I'll defer to @cprice404 about the README template choices.

README.md
nand4011: I did it manually since we had trouble getting the readme updated before and the code has changed enough that I didn't want the example in the readme showing the old code.

cprice404 (approved):

PR #296, opened 2023-06-29
chore: Renamed lib and moved READMe to have a sub directory structure for java examples

pratik151192:
There are no functional changes in this PR. I have simply moved all the code from "lib" to "cache-lib".


pratik151192:
Closing as CI is failing due to moving directories as our dev docs references it. Discussing with team if we want to do a 3 phased commit to make sure docs don't break or simply live with lib as a directory name and create a new one with the aws dependencies/code

pratik151192:
Re-opening a new one as this has become a bit clumsy after a few changes post discussion with Chris

PR #297, opened 2023-06-29
chore: Renamed lib and moved READMe to have a sub directory structure for java examples

pratik151192:
### Description

This commit doesn't involve any functional change but only moves the files around from "lib" directory to a directory called "cache". We are doing this to add examples for Momento with AWS Secrets Manager and the idea is to have that in it's own directory with it's own gradle file so that we aren't unnecessarily pulling all dependencies for every example.

### Testing

```
➜  examples git:(rename-lib) MOMENTO_AUTH_TOKEN="FAKE_TOKEN" ./gradlew basic

> Task :cache:basic
******************************************************************
Basic Example Start
******************************************************************
16:44:45.159 [main] INFO  momento.sdk.CacheClient -- Creating Momento Cache Client
Cache with name 'cache' already exists.
Listing caches:
default-cache
cache
Setting key 'key', value 'value'
Getting value for key 'key'
Found value for key 'key': 'value'
******************************************************************
Basic Example End
******************************************************************

BUILD SUCCESSFUL in 1s
3 actionable tasks: 1 executed, 2 up-to-date
```


cprice404 (approved):
lovely. thanks for doing this as a separate PR!

PR #300, opened 2023-07-07
chore: add readme for top level example directory

pratik151192:


.github/dependabot.yml
```diff
@@ -6,8 +6,15 @@
...
       interval: "daily"
     allow:
       - dependency-name: "software.momento.java:sdk"
+  - package-ecosystem: "gradle" # See documentation for possible values
+    directory: "/examples/cache-with-aws" # Location of package manifests
+    schedule:
+      interval: "daily"
+    allow:
+      - dependency-name: "software.momento.java:sdk"
+      - dependency-name: "software.amazon.awssdk:secretsmanager"
```
pratik151192, line 20: I'm not 100% if this will be on a new line and couldn't find examples or reference to check.

cprice404: for now let's just only do the momento dep. the purpose here is to test our released packages, we can upgrade other deps more intentionally over time, we don't necessarily need them auto-updated.

examples/README.md
```diff
@@ -0,0 +1,10 @@
+# Java Client SDK
+
+## Examples
+
+All examples are in subdirectories of `examples/cache` and `examples/cache-with-aws`. We have different directories for different examples in order to keep the `build.gradle.kts` files to a minimum for users who may be copying things from them. For instance, when we have examples that use the AWS SDK, we don't want users to think they need those deps for a basic momento project, so we put them in their own directory with their own ```build.gradle.kts```.
```
cprice404, line 5: I think you can omit this paragraph for now, this was more intended for internal info for our devs

cprice404 (approved):

PR #303, opened 2023-07-21
feat: add retry interceptor for JAVA SDK

pratik151192:
### **Note**: 

Apart from adding retries, this PR contains backward incompatible changes with respect to cache client initialization. We made this decision as the SDK is relatively young and to follow better and consistent practices with other SDKs. Previously, if a user wanted to override the gRPC request deadline, they did:

```
  CacheClient.builder(
                        credentialProvider, Configurations.Laptop.latest(), DEFAULT_TTL_SECONDS)
                    .setDeadline(Duration.ofMillis(0))
                    .build());
```

The `setDeadline` method violates the builder pattern and overrides an already initialized configuration. We also want to remove the word deadline as it's not conventional and gRPC specific. This PR removes `setDeadline` from the builder, and adds a `withTimeout` to the Configuration object which returns a new Configuration object with the appropriate gRPC transport strategy based on the timeout/deadline. The initialization now looks like

```
  CacheClient.builder(
                        credentialProvider,
                        Configurations.Laptop.latest()
                                    .withTimeout(Duration.ofMillis(100)),
                        DEFAULT_TTL_SECONDS)
                    .build());
```

### Overview 

Added a ```RetryInterceptor``` to intercept gRPC requests to add SDK/client-side retries. I have added enough JavaDocs on the code to make sense of what's going on, but please call out if you want more details here or over a call. Note that this hasn't been wrapped as a middleware, but there's nothing stopping us in future if we want to do so.

There are 2 new constructs in this PR from a client perspective:

- **Retry Strategy**: This interface contains a contract ```determineWhenToRetry``` (optional) which can be overridden by clients to provide their own retry strategy. Otherwise, we default to our own flavor of ```FixedCountRetryStrategy``` that simply retries upto 3 times without any delay. The PR also adds a ```FixedDelayRetryStrategy``` that clients can directly use if they want to, though that's not the default.
- **Retry Eligbility Strategy**: This is composed int the RetryStrategy class and a default is available.```DefaultRetryEligibiliyStrategy``` kicks in if clients don't provide the eligibility, which checks for eligible status codes and method names to retry on.
 
The code is backwards compatible from a client initialization perspective, so if they do

```
      CacheClient.builder(credentialProvider, Configurations.Laptop.latest(), DEFAULT_TTL_SECONDS)
            .build();
```

like before, they get the retry and eligiblity strategies for free for the default. They can override both like:

```
      RetryStrategy retryStrategy = new FixedCountRetryStrategy(10);
      CacheClient.builder(credentialProvider, Configurations.Laptop.latest().withRetryStrategy(retryStrategy), DEFAULT_TTL_SECONDS)
            .build();
```

OR

```
      RetryEligibilityStrategy eligibilityStrategy = new CustomEligibilityStrategy();
      RetryStrategy retryStrategy = new FixedCountRetryStrategy(10, eligibilityStrategy);

      CacheClient.builder(credentialProvider, Configurations.Laptop.latest().withRetryStrategy(retryStrategy), DEFAULT_TTL_SECONDS)
            .build();
```

### Testing

- I have manually tested that retry kicks in by tinkering with the status codes and method names; to check if the retry logic kicks in on an authentication or cache doesn't exist error. It's hard to reproduce a timeout/network error so I will run a loadgen against this.
- Will do some more custom testing but sending out PR for feedback.

References: https://github.com/facebook/buck/blob/main/src/com/facebook/buck/remoteexecution/grpc/retry/RetryClientInterceptor.java

### DIFF
--- a/momento-sdk/build.gradle.kts
+++ b/momento-sdk/build.gradle.kts
@@ -12,6 +12,7 @@ dependencies {
     api(libs.grpc.api) // Marked api because SdkException contains classes from this dependency
     implementation(libs.grpc.stub)
     implementation(libs.grpc.nettyshaded)
+    implementation(libs.grpc.context)
     implementation(libs.protobuf.java)
     implementation(libs.guava)
     implementation(libs.gson)
@@ -25,6 +26,9 @@ dependencies {
     // Test dependencies
     testImplementation(libs.junit)
     testImplementation(libs.assertj)
+    testImplementation(libs.slf4j.api)
+    testImplementation(libs.logback)
+
 }
 
 spotless {
--- a/momento-sdk/build.gradle.kts
+++ b/momento-sdk/build.gradle.kts
@@ -26,9 +26,10 @@ dependencies {
     // Test dependencies
     testImplementation(libs.junit)
     testImplementation(libs.assertj)
+    testImplementation(libs.mockito)
+    testImplementation(libs.mockito.jupiter)
     testImplementation(libs.slf4j.api)
     testImplementation(libs.logback)
-
 }
 
 spotless {
--- a/momento-sdk/src/main/java/momento/sdk/CacheClientBuilder.java
+++ b/momento-sdk/src/main/java/momento/sdk/CacheClientBuilder.java
@@ -4,12 +4,6 @@ import java.time.Duration;
 import javax.annotation.Nonnull;
 import momento.sdk.auth.CredentialProvider;
 import momento.sdk.config.Configuration;
-import momento.sdk.config.transport.GrpcConfiguration;
-import momento.sdk.config.transport.TransportStrategy;
-import momento.sdk.retry.DefaultRetryEligibilityStrategy;
-import momento.sdk.retry.FixedCountRetryStrategy;
-import momento.sdk.retry.RetryEligibilityStrategy;
-import momento.sdk.retry.RetryStrategy;
 
 /** Builder for {@link CacheClient} */
 public final class CacheClientBuilder {
@@ -33,40 +27,6 @@ public final class CacheClientBuilder {
     this.configuration = configuration;
     ValidationUtils.ensureValidTtl(itemDefaultTtl);
     this.itemDefaultTtl = itemDefaultTtl;
-  }
-
-  /**
-   * Sets the maximum duration of a client call.
-   *
-   * @param deadline The deadline duration.
-   * @return The updated builder.
-   */
-  public CacheClientBuilder setDeadline(@Nonnull Duration deadline) {
-    ValidationUtils.ensureRequestDeadlineValid(deadline);
-
-    final GrpcConfiguration newGrpcConfiguration =
-        configuration.getTransportStrategy().getGrpcConfiguration().withDeadline(deadline);
-    final TransportStrategy newTransportStrategy =
-        configuration.getTransportStrategy().withGrpcConfiguration(newGrpcConfiguration);
-    final RetryStrategy retryStrategy =
-        configuration.getRetryStrategy() == null
-            ?
-            // create default if the configuration didn't have a retry strategy otherwise don't
-            // override a customer provided strategy
-            new FixedCountRetryStrategy(Configuration.MAX_RETRIES)
-            : configuration.getRetryStrategy();
-    final RetryEligibilityStrategy retryEligibilityStrategy =
-        configuration.getRetryEligibilityStrategy() == null
-            ?
-            // create default if the configuration didn't have a retry eligbility strategy otherwise
-            // don't override a customer provided strategy
-            new DefaultRetryEligibilityStrategy()
-            : configuration.getRetryEligibilityStrategy();
-    configuration =
-        configuration.withTransportStrategy(
-            newTransportStrategy, retryStrategy, retryEligibilityStrategy);
-
-    return this;
   }
 
   /**
--- a/momento-sdk/src/main/java/momento/sdk/CacheClientBuilder.java
+++ b/momento-sdk/src/main/java/momento/sdk/CacheClientBuilder.java
@@ -6,6 +6,10 @@ import momento.sdk.auth.CredentialProvider;
 import momento.sdk.config.Configuration;
 import momento.sdk.config.transport.GrpcConfiguration;
 import momento.sdk.config.transport.TransportStrategy;
+import momento.sdk.retry.DefaultRetryEligibilityStrategy;
+import momento.sdk.retry.FixedCountRetryStrategy;
+import momento.sdk.retry.RetryEligibilityStrategy;
+import momento.sdk.retry.RetryStrategy;
 
 /** Builder for {@link CacheClient} */
 public final class CacheClientBuilder {
@@ -44,7 +48,23 @@ public final class CacheClientBuilder {
         configuration.getTransportStrategy().getGrpcConfiguration().withDeadline(deadline);
     final TransportStrategy newTransportStrategy =
         configuration.getTransportStrategy().withGrpcConfiguration(newGrpcConfiguration);
-    configuration = configuration.withTransportStrategy(newTransportStrategy);
+    final RetryStrategy retryStrategy =
+        configuration.getRetryStrategy() == null
+            ?
+            // create default if the configuration didn't have a retry strategy otherwise don't
+            // override a customer provided strategy
+            new FixedCountRetryStrategy(Configuration.MAX_RETRIES)
+            : configuration.getRetryStrategy();
+    final RetryEligibilityStrategy retryEligibilityStrategy =
+        configuration.getRetryEligibilityStrategy() == null
+            ?
+            // create default if the configuration didn't have a retry eligbility strategy otherwise
+            // don't override a customer provided strategy
+            new DefaultRetryEligibilityStrategy()
+            : configuration.getRetryEligibilityStrategy();
+    configuration =
+        configuration.withTransportStrategy(
+            newTransportStrategy, retryStrategy, retryEligibilityStrategy);
 
     return this;
   }
--- a/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
+++ b/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
@@ -11,7 +11,11 @@ import io.grpc.MethodDescriptor;
 import io.grpc.Status;
 import java.time.Duration;
 import java.util.Optional;
-import java.util.concurrent.*;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
 import javax.annotation.Nullable;
 import momento.sdk.retry.RetryEligibilityStrategy;
 import momento.sdk.retry.RetryStrategy;
--- a/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
+++ b/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
@@ -1,0 +1,189 @@
+package momento.sdk;
+
+import com.google.common.util.concurrent.MoreExecutors;
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.Context;
+import io.grpc.ForwardingClientCallListener;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+import javax.annotation.Nullable;
+import momento.sdk.retry.RetryEligibilityStrategy;
+import momento.sdk.retry.RetryStrategy;
+import momento.sdk.retry.RetryingUnaryClientCall;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Interceptor for retrying client calls with gRPC servers. This interceptor is responsible for
+ * handling retry logic when making unary (single request, single response) gRPC calls.
+ *
+ * <p>A {@link ClientCall} is essentially an instance of a gRPC invoker. Every gRPC interceptor
+ * expects us to return such client call(s) that it will execute in order. Each call has a "start"
+ * method, which is the entry point for the call.
+ *
+ * <p>This retry client interceptor returns an instance of a {@link RetryingUnaryClientCall}, which
+ * is a client call designed to handle retrying unary (single request, single response) operations.
+ * The interceptor uses a provided {@link RetryStrategy} and {@link RetryEligibilityStrategy} to
+ * determine when and how to retry failed calls.
+ *
+ * <p>The interceptor is constructed with a {@link RetryStrategy} and a {@link
+ * RetryEligibilityStrategy}, both of which can be customized based on specific requirements. The
+ * {@link RetryStrategy} is responsible for providing the delay between retry attempts, while the
+ * {@link RetryEligibilityStrategy} helps decide whether a request is eligible for retry based on
+ * the status and method details.
+ *
+ * <p>When a gRPC call is intercepted, the interceptor checks whether the method is unary (client
+ * sends one message), and if so, it wraps the original {@link ClientCall} with the {@link
+ * RetryingUnaryClientCall}. This custom call is responsible for handling the retry logic.
+ *
+ * <p>When the gRPC call is closed, the {@code onClose} method is called, which is the point where
+ * we can safely check the status of the initial request that was made and determine if we want to
+ * retry or not. If the request was successful or ineligible for retry (based on the {@link
+ * RetryEligibilityStrategy}), the interceptor returns the result to the original listener,
+ * effectively completing the call. Otherwise, it starts the retrying process by scheduling a new
+ * call attempt with a delay provided by the {@link RetryStrategy}.
+ *
+ * <p>If the retry attempts are exhausted or if the provided delay is not present (indicating that
+ * we should not retry anymore), the interceptor propagates the final result to the original
+ * listener, effectively completing the call with the last status received.
+ *
+ * <p>Note that the interceptor only supports unary operations for retrying.
+ *
+ * @see RetryStrategy
+ * @see RetryEligibilityStrategy
+ * @param <ReqT> The type of the request message.
+ * @param <RespT> The type of the response message.
+ */
+final class RetryClientInterceptor implements ClientInterceptor {
+
+  private final RetryStrategy retryStrategy;
+  private final RetryEligibilityStrategy retryEligibilityStrategy;
+  private final Logger logger = LoggerFactory.getLogger(RetryClientInterceptor.class);
+
+  public RetryClientInterceptor(
+      final RetryStrategy retryStrategy, final RetryEligibilityStrategy retryEligibilityStrategy) {
+    this.retryStrategy = retryStrategy;
+    this.retryEligibilityStrategy = retryEligibilityStrategy;
+  }
+
+  @Override
+  public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+      final MethodDescriptor<ReqT, RespT> method,
+      final CallOptions callOptions,
+      final Channel channel) {
+    // currently the SDK only supports unary operations which we want to retry on
+    if (!method.getType().clientSendsOneMessage()) {
+      return channel.newCall(method, callOptions);
+    }
+
+    return new RetryingUnaryClientCall<ReqT, RespT>(channel.newCall(method, callOptions)) {
+      private int attemptNumber = 0;
+      @Nullable private Future<?> future = null;
+
+      @Override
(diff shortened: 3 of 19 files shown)


momento-sdk/src/main/java/momento/sdk/CacheClientBuilder.java
```diff
@@ -44,7 +48,23 @@ public CacheClientBuilder setDeadline(@Nonnull Duration deadline) {
         configuration.getTransportStrategy().getGrpcConfiguration().withDeadline(deadline);
     final TransportStrategy newTransportStrategy =
         configuration.getTransportStrategy().withGrpcConfiguration(newGrpcConfiguration);
-    configuration = configuration.withTransportStrategy(newTransportStrategy);
```
pratik151192, line 47: I understand why we chose to give a ```setDeadline``` option to clients to make them decoupled with our gRPC logic. However, this pattern overrides the configuration object, and needed the below logic to handle not overriding other strategies the client provides. This has a weird anti-pattern effect

pratik151192: It seemed an anti pattern to me because the Builder gives an option to setDeadline, which inadvertently overrides the configuration with a new transport strategy. Therefore, if we add more strategies or other things to the configuration, we always need to make sure we don't override the client provided settings in this method. This is already consumed internally, and maybe consumed externally, so i didn't refactor this part. Ideally, the configuration should not be overriden and just set the deadline for the existing strategy.

momento-sdk/src/main/java/momento/sdk/retry/FixedCountRetryStrategy.java
```diff
@@ -0,0 +1,42 @@
...
+   */
+  public FixedCountRetryStrategy(final int maxAttempts) {
+    Preconditions.checkArgument(
+        maxAttempts > 0, "Total number " + "of retry attempts should be greater than 0");
+    this.maxAttempts = maxAttempts;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public Optional<Long> getDelay(final int currentAttempt) {
```
cprice404, line 36: you probably saw that in the other sdks we call this `determineWhenToRetry` or something.

Your name is totally reasonable and in a green field I'd definitely have been open to discussion between those two names.

But given that some of our other SDKs are more mature and widely used than this one I think there's at least a small argument to be made that consistency would be valuable.

I don't feel super strongly about it, it would only ever be relevant to a polyglot developer, and on top of that someone who needed an advanced/custom retry strategy, and even then it's probably not hard for someone to figure it out.

So I'm happy to let you make the choice whether the consistency is worthwhile here.

momento-sdk/src/main/java/momento/sdk/retry/RetryStrategy.java
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Retrieves the delay in milliseconds for the next retry attempt based on the current attempt
+   * number.
+   *
+   * @param currentAttempt The current attempt number of the retry. Starts from 1 for the first
+   *     retry attempt.
+   * @return An {@link Optional} containing the delay in milliseconds for the next retry attempt, or
+   *     an empty optional to indicate that no further retry attempts should be made.
+   */
+  Optional<Long> getDelay(int currentAttempt);
```
cprice404, line 43: What if a really advanced user wanted to write a strategy that would retry `set` requests 5 times but `get` requests only 2 times? Or they were willing to retry `UNAVAILABLE` more times than some other status code? Or retry UNAVAILABLE with a more aggressive delay than some other error codes?

momento-sdk/src/main/java/momento/sdk/config/Configuration.java
```diff
@@ -25,6 +51,26 @@ public TransportStrategy getTransportStrategy() {
...
+
+  /**
+   * Strategy to determine if a request is eligible for retry. By default, {@link
+   * DefaultRetryEligibilityStrategy} gets used.
+   *
+   * @return The retry eligibility strategy
+   */
+  public RetryEligibilityStrategy getRetryEligibilityStrategy() {
+    return retryEligibilityStrategy;
+  }
```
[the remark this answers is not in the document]
cprice404: actually after I looked at the interfaces more closely I realized there might be some use cases that this separation would preclude in the future.  See my comment in the RetryStrategy.java file.

cprice404 (commented):
nice work man! handful of questions/nits but looking really good.

momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
```diff
@@ -0,0 +1,189 @@
...
+                // delay or any variable the strategy author wishes to not retry anymore
+                if (!retryDelay.isPresent()) {
+                  super.onClose(status, trailers);
+                  return;
+                }
+
+                // Thread.sleep(0) also makes the OS prioritize it at some point
+                if (retryDelay.get() > 0L) {
+                  try {
+                    Thread.sleep(retryDelay.get());
```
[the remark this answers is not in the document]
pratik151192: Ideally we use a ```ScheduledExecutorService``` and let JVM handle the scheduling part without doing any sort of "sleeping". You're right that it's not ideal to use sleep, and in this case I can use a simple wait()-notify() mechanism to achieve the wait behavior. wait() releases the lock on the current thread, whereas sleep() doesn't so will be more performant, but not as precise as sleep(). I actually did think of going over this part again as I agree with your concerns. Will look more!

momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
```diff
@@ -0,0 +1,189 @@
...
+                  } catch (InterruptedException e) {
+                    // Restore the interrupted status
+                    Thread.currentThread().interrupt();
```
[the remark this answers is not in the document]
pratik151192: It's a pretty standard practice in Java that if you the thread gets interrupted, we reset the interrupted flag. By restoring the interrupted status, we are informing the calling code (higher up the call stack) that the thread was interrupted, and it should respond appropriately. It allows other parts of the application to handle the interruption as needed. This is not applicable here in theory I know, but it's a safe practice to follow.

momento-sdk/src/main/java/momento/sdk/CacheClientBuilder.java
pratik151192: I agree with all the three and 100% agreed on violating the builder pattern. I was skeptical of removing this part to not break backward compatibility, but if we are young then we might take the bullet and do it as this was cause problems in the future as well.

momento-sdk/src/main/java/momento/sdk/CacheClientBuilder.java
```diff
@@ -44,7 +48,23 @@ public CacheClientBuilder setDeadline(@Nonnull Duration deadline) {
...
+    final RetryEligibilityStrategy retryEligibilityStrategy =
+        configuration.getRetryEligibilityStrategy() == null
+            ?
+            // create default if the configuration didn't have a retry eligbility strategy otherwise
+            // don't override a customer provided strategy
+            new DefaultRetryEligibilityStrategy()
+            : configuration.getRetryEligibilityStrategy();
+    configuration =
+        configuration.withTransportStrategy(
+            newTransportStrategy, retryStrategy, retryEligibilityStrategy);
```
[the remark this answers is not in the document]
pratik151192: Will get red of this totally and make it comply the builder pattern

momento-sdk/src/main/java/momento/sdk/config/Configuration.java
```diff
@@ -25,6 +51,26 @@ public TransportStrategy getTransportStrategy() {
     return transportStrategy;
   }
 
+  /**
+   * Configuration for retry tunables. By default, {@link momento.sdk.retry.FixedDelayRetryStrategy}
```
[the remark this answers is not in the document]
pratik151192: good catch. will change!

momento-sdk/src/main/java/momento/sdk/config/Configuration.java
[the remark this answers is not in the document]
pratik151192: Love the composition idea of a ```RetryConfig```

momento-sdk/src/main/java/momento/sdk/config/Configurations.java
```diff
@@ -39,7 +46,40 @@ public static Configuration latest() {
...
+    }
+
+    /**
+     * Provides v1 configuration for a laptop environment with a custom {@link RetryStrategy}. This
+     * configuration is guaranteed not to change in future releases of the Momento Java SDK.
+     *
+     * @param retryStrategy the retry strategy
+     * @return the v1 laptop configuration
+     */
+    public static Configuration v1(RetryStrategy retryStrategy) {
```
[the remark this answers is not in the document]
pratik151192: +1 to the copy constructor.. will do that!

momento-sdk/src/main/java/momento/sdk/retry/FixedCountRetryStrategy.java
pratik151192: I dont have a strong opinion either :) I will make it consistent as that sounds something we should do!

momento-sdk/src/main/java/momento/sdk/retry/RetryStrategy.java
pratik151192: I'll rethink the design based on this use-case and the other comment on ```Configuration.java```

momento-sdk/build.gradle.kts
```diff
@@ -25,6 +26,9 @@ dependencies {
     // Test dependencies
     testImplementation(libs.junit)
     testImplementation(libs.assertj)
+    testImplementation(libs.slf4j.api)
```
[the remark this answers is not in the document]
cprice404: oh, okay then I don't think you need to add it as a test dep, it will already be available from the main implementation deps?

should be harmless especially since y'all consolidated the version numbers / artifacts to a single place, so nbd if you have a reason you want to keep it.

momento-sdk/src/main/java/momento/sdk/ScsDataGrpcStubsManager.java
```diff
@@ -8,7 +8,7 @@
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
cprice404, line 11: nit: we generally try to avoid wildcard imports. you can change the settings in intellij to make it not do this. I don't feel too strongly about it though.

momento-sdk/src/main/java/momento/sdk/ScsDataGrpcStubsManager.java
```diff
@@ -25,22 +25,65 @@ final class ScsDataGrpcStubsManager implements Closeable {
   private final ManagedChannel channel;
   private final ScsGrpc.ScsFutureStub futureStub;
   private final Duration deadline;
+  private final ScheduledExecutorService retryScheduler;
+  private final ExecutorService retryExecutor;
+
+  // An arbitary selection of twice the number of available processors.
+  private static final int MAX_RETRY_THREAD_POOL_SIZE =
```
cprice404, line 32: I'm not sure it makes sense to bound this based on the number of processors? That would definitely make sense for a cpu-bound thread pool but this is an IO-bound thread pool.

In Kotlin they default it to 64: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-i-o.html

momento-sdk/src/main/java/momento/sdk/ScsDataGrpcStubsManager.java
```diff
@@ -25,22 +25,65 @@ final class ScsDataGrpcStubsManager implements Closeable {
...
+     * dynamically.
+     *
+     * <p>The {@link ThreadPoolExecutor} has a size of 0 and grows onDemand. The maximumPoolSize is
+     * present to cap the number of threads we create for retries; where the {@link
+     * LinkedBlockingQueue} essentially stores any pending retries. Note that the keep alive time
+     * for each thread is 60 seconds, beyond which the JVM will destroy the thread. This is the
+     * default value of {@link Executors.newCachedThreadPool()} and also applies in our situation as
+     * even with backoffs, one retry should be below 60 seconds. We aren't directly using a cached
+     * thread pool because it grows unbounded.
+     */
```
cprice404, line 62: What would you expect the difference in behavior to be if we just created one executor (the one you are currently calling `retryExecutor`) and gave it a minimum size of 1, and used it for both tasks?

cprice404 (commented):
handful of small questions/nits remaining but I'm pretty much +1 pending your responses and @nand4011 's approval.

momento-sdk/src/main/java/momento/sdk/ScsDataGrpcStubsManager.java
pratik151192: Ah my IDE is new on a new machine!!!! I feel strongly about it :) will kill this!

momento-sdk/src/main/java/momento/sdk/ScsDataGrpcStubsManager.java
pratik151192: Good point. No opinion on the number in general so use 64

momento-sdk/src/main/java/momento/sdk/ScsDataGrpcStubsManager.java
pratik151192: The ```retryExecutor``` is a ```ThreadPoolExecutor``` and doesn't provide scheduling semantics. It runs tasks on-demand. The ```retryScheduler``` is a ```ScheduledThreadPoolExecutor``` that schedules tasks with a delay. We could use a ```ScheduledThreadPoolExecutor``` to do both, but it creates idle threads and not on-demand. If we used only this for execution with a size of 1, then all retries will happen sequentially.

pratik151192:
@cprice404 responded.. will cut a new revision once Nate reviews as your comments are relatively small changes :)

momento-sdk/src/main/java/momento/sdk/ScsDataGrpcStubsManager.java
cprice404: does the `retryScheduler` sleep in order to achieve the delay? or does it just immediately schedule the task on another thread and add in a sleep to that task that it is scheduling?

Just want to make sure that:
(a) the retryScheduler isn't going to be blocked while we are waiting for the delay to transpire (because if it was then we could only schedule one retry at a time), and
(b) it is worth the (minor) extra complexity of having two thread pools to reason about instead of one.

cprice404 (commented):
if you'd like me to review again after nate does feel free to re-request review at that time, but at this point I'm pretty sure I'm +1 once you two are happy.

momento-sdk/src/main/java/momento/sdk/ScsDataGrpcStubsManager.java
pratik151192: The ```retryScheduler``` does not use sleep internally. Instead, it relies on a combination of a task queue, a thread pool, and a DelayQueue (a priority queue of delayed tasks) to handle scheduling and execution of tasks. I have actually implemented this executor from scratch around a couple years back, we can discuss this more if you'd like in our 1:1 :)

pratik151192: The scheduling does happen on a single thread though, but in this case the scheduler is just submitting the task to another executor and moving onto the next Runnable to schedule. It doesn't block but wakes up when there's a task in in's internal queue, and then checks if the delay has elapsed (using epoch time). This check might happen for a bunch of retries in parallel, but again, it's not doing the real work. We can always increase this number though I believe it wouldn't be necessary; I can do a small toy benchmarking if you'd like.

momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  @Override
+  public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+      final MethodDescriptor<ReqT, RespT> method,
+      final CallOptions callOptions,
+      final Channel channel) {
+    // currently the SDK only supports unary operations which we want to retry on
+    if (!method.getType().clientSendsOneMessage()) {
+      return channel.newCall(method, callOptions);
```
nand4011, line 78: Will this be intercepted again?

nand4011, line 78: Will this be intercepted again?

momento-sdk/src/main/java/momento/sdk/config/Configurations.java
```diff
@@ -4,18 +4,21 @@
 import momento.sdk.config.transport.GrpcConfiguration;
 import momento.sdk.config.transport.StaticTransportStrategy;
 import momento.sdk.config.transport.TransportStrategy;
+import momento.sdk.retry.*;
```
nand4011, line 7: I'm sure you'll find and replace these, so I won't point out each of them.

momento-sdk/src/main/java/momento/sdk/retry/DefaultRetryEligibilityStrategy.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private static final Set<Status.Code> RETRYABLE_STATUS_CODES = new HashSet<>();
+
+  static {
+    RETRYABLE_STATUS_CODES.add(Status.Code.UNAVAILABLE);
+    RETRYABLE_STATUS_CODES.add(Status.Code.INTERNAL);
+  }
+
+  private static final Set<String> RETRYABLE_GRPC_FULL_METHOD_NAMES =
+      new HashSet<String>() {
```
nand4011, line 22: We should have the sorted set methods represented here, unless none of them are retryable.

momento-sdk/src/main/java/momento/sdk/retry/FixedCountRetryStrategy.java
```diff
@@ -0,0 +1,69 @@
...
+   * @param eligibilityStrategy a strategy that determines if the gRPC status code and methods are
+   *     eligible or safe to retry.
+   * @throws IllegalArgumentException if maxAttempts is not greater than 0.
+   */
+  public FixedCountRetryStrategy(
+      final int maxAttempts, final RetryEligibilityStrategy eligibilityStrategy) {
+    Preconditions.checkArgument(
+        maxAttempts > 0, "Total number " + "of retry attempts should be greater than 0");
+    Preconditions.checkNotNull(
+        eligibilityStrategy, "Retry eligibility strategy should" + " not be null");
```
nand4011, line 39: Does the + do anything here?

momento-sdk/src/main/java/momento/sdk/retry/FixedCountRetryStrategy.java
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * @param maxAttempts The maximum number of retry attempts. After reaching this limit, no more
+   *     retries will be performed, and the strategy will return an empty optional.
+   * @param eligibilityStrategy a strategy that determines if the gRPC status code and methods are
+   *     eligible or safe to retry.
+   * @throws IllegalArgumentException if maxAttempts is not greater than 0.
+   */
+  public FixedCountRetryStrategy(
+      final int maxAttempts, final RetryEligibilityStrategy eligibilityStrategy) {
+    Preconditions.checkArgument(
```
nand4011, line 36: Did we decide we were alright with expanding our use of Guava?

nand4011 (commented):
LGTM other than a couple of formatting things and the sorted set methods in the list of retryable methods.

momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
pratik151192: This is the client call which doesn't return our interceptor if it's a non-unary operation. The unary operations are the ones that provide the RetryingClientCall and those will be intercepted. I did a quick test on my end where I added ```NOT_FOUND``` as a retryable status and put my cache name as a random string. I enabled the log and saw 3 attempts being made.

momento-sdk/src/main/java/momento/sdk/retry/FixedCountRetryStrategy.java
pratik151192: nah will remove

momento-sdk/src/main/java/momento/sdk/retry/FixedCountRetryStrategy.java
pratik151192: I'm not sure.. but good point I'll remove the use of it to not keep expanding and piling rework later

momento-sdk/src/main/java/momento/sdk/ScsDataGrpcStubsManager.java
cprice404: it's an implementation detail, we can change it in the future, I trust your judgment if you are confident in the semantics of it and if you believe this is the simplest way to implement it.

cprice404 (approved):

PR #304, opened 2023-07-24
feat: Add a load generator to the examples

nand4011:
Add a load generator example in the style of load generators we've made for other SDKs. It schedules a configurable number of alternating write and read tasks, limiting the total tps using a Guava RateLimiter.

### DIFF
--- a/examples/cache/build.gradle.kts
+++ b/examples/cache/build.gradle.kts
@@ -19,8 +19,13 @@ repositories {
 dependencies {
     implementation("software.momento.java:sdk:1.0.0")
 
+    implementation("com.google.guava:guava:31.1-android")
+
     // Logging framework to log and enable logging in the Momento client.
     implementation("ch.qos.logback:logback-classic:1.4.7")
+
+    // Histogram for collecting stats in the load generator
+    implementation("org.hdrhistogram:HdrHistogram:2.1.12")
 
     // Use JUnit Jupiter for testing.
     testImplementation("org.junit.jupiter:junit-jupiter:5.9.2")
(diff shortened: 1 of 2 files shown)


examples/cache/build.gradle.kts
```diff
@@ -19,9 +19,14 @@ repositories {
 dependencies {
     implementation("software.momento.java:sdk:1.0.0")
 
+    implementation("com.google.guava:guava:31.1-android")
```
nand4011, line 22: This is the version of guava transitively pulled in by our protos.

examples/cache/src/main/java/momento/client/example/LoadGenerator.java
```diff
@@ -0,0 +1,262 @@
...
+        workerId,
+        operationNum,
+        key -> client.get(CACHE_NAME, key),
+        (response, operationNumValue) -> {
+          if (response instanceof GetResponse.Hit || response instanceof GetResponse.Miss) {
+            globalSuccessCount.increment();
+          } else if (response instanceof GetResponse.Error error) {
+            handleErrorResponse(error.getErrorCode());
+          }
+          scheduleSet(workerId, nextOperationNum);
```
pratik151192, line 112: This looks like unintended else we'll be in an infinite loop of set-get-set?

pratik151192 (commented):
Just had 1 comment otherwise LGTM

examples/cache/src/main/java/momento/client/example/LoadGenerator.java
nand4011: That is intended. It schedules alternating write and read tasks that reschedule themselves forever. The throughput is controlled by the rate limiter blocking when the tasks run too quickly. It is broken by the shutdown method that the main method calls at the end of the benchmark.

pratik151192 (approved):

PR #313, opened 2023-08-22
feat: add factory method that eagerly connects with lambda examples

pratik151192:
This PR adds the following to Java SDK

- Eager connection for both sync client through a new factory method
- Lambda configuration
- Example for Java SDK and AWS Lambda
   - This will be updated to use the Lambda configuration created above in the next PR

### Testing

```

16:30:55.501 [grpc-default-executor-1] INFO  momento.sdk.ScsDataGrpcStubsManager -- State transitioned to CONNECTING; waiting to get READY
16:30:55.815 [grpc-default-executor-0] INFO  momento.sdk.ScsDataGrpcStubsManager -- Connected to Momento's server! Happy Caching!

CacheDataPlaneEagerConnectionTest > getReturnsHitAfterSet() PASSED
BUILD SUCCESSFUL in 3s

```

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/CacheDataPlaneEagerConnectionTest.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/CacheDataPlaneEagerConnectionTest.java
@@ -8,9 +8,10 @@ import java.time.Duration;
 import momento.sdk.config.Configurations;
 import momento.sdk.responses.cache.GetResponse;
 import momento.sdk.responses.cache.SetResponse;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
-public class CacheDataPlaneEagerConnectionTest {
+public class CacheDataPlaneEagerConnectionTest extends BaseTestClass {
   private static final Duration DEFAULT_ITEM_TTL_SECONDS = Duration.ofSeconds(60);
   private final String cacheName = System.getenv("TEST_CACHE_NAME");
 
--- a/momento-sdk/src/intTest/java/momento/sdk/CacheDataPlaneEagerConnectionTest.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/CacheDataPlaneEagerConnectionTest.java
@@ -1,0 +1,39 @@
+package momento.sdk;
+
+import static momento.sdk.BaseTestClass.credentialProvider;
+import static momento.sdk.TestUtils.randomString;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.time.Duration;
+import momento.sdk.config.Configurations;
+import momento.sdk.responses.cache.GetResponse;
+import momento.sdk.responses.cache.SetResponse;
+import org.junit.jupiter.api.Test;
+
+public class CacheDataPlaneEagerConnectionTest {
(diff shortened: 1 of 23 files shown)


momento-sdk/src/intTest/java/momento/sdk/CacheDataPlaneEagerConnectionTest.java
```diff
@@ -0,0 +1,39 @@
...
+            DEFAULT_ITEM_TTL_SECONDS,
+            Duration.ofSeconds(10));
+    final String key = randomString("key");
+    final String value = randomString("value");
+
+    // Successful Set
+    final SetResponse setResponse = client.set(cacheName, key, value).join();
+    assertThat(setResponse).isInstanceOf(SetResponse.Success.class);
+    assertThat(((SetResponse.Success) setResponse).valueString()).isEqualTo(value);
+    assertThat(((SetResponse.Success) setResponse).valueByteArray()).isEqualTo(value.getBytes());
```
cprice404, line 32: this is more just a forward-thinking thing, but i wonder what it would take for us to upgrade the tests to use JDK 17 so that we can use JEP 394 in the tests? we obviously need to make sure the SDK itself can be used with JDKs at least back to 11, but for our own tests it sure would be nice to be able to use the modern patterns.

cprice404 (commented):
just a couple of minor wording nits, and would be good for @nand4011 to review

momento-sdk/src/intTest/java/momento/sdk/CacheDataPlaneEagerConnectionTest.java
nand4011: I'm in favor of this if it's possible to have different language levels for the sdk and tests without having them in separate modules or something like that.

momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -85,6 +88,48 @@ public CacheClient(
...
+    // eager connection
+    if (timeout != 0) {
+      client.scsDataClient.connect(timeout);
+    }
+
+    return client;
+  }
+
+  private static void checkTimeoutValid(long timeout) {
+    if (timeout < 0) {
```
nand4011, line 128: I feel like a negative timeout can reasonably be treated as infinite, treated as 0, or throw. I'd lean slightly against throwing since the method can still return an unconnected client in the worst case. I think it's a minor thing overall though.

nand4011 (commented):
Looks good after Chris' suggestions.

momento-sdk/src/intTest/java/momento/sdk/CacheDataPlaneEagerConnectionTest.java
pratik151192: For the interest of time, I have created an issue for this and will be happy to fast follow https://github.com/momentohq/client-sdk-java/issues/314. I agree with both of your sentiments though.

momento-sdk/src/main/java/momento/sdk/CacheClient.java
pratik151192: I agree with you. The only reason I'll keep it as is for now is because we have done the same in 3 other SDKs.

nand4011 (approved):

PR #317, opened 2023-08-28
chore: overload create method with null timeout

pratik151192:


nand4011 (commented):
I'm trying to make up my mind about whether this method should not eagerly connect, or eagerly connect with the default timeout. I'm leaning slightly toward connecting, because people are going to use this in a lambda without thinking about it. I think we need more input though.

cprice404:
> I'm trying to make up my mind about whether this method should not eagerly connect, or eagerly connect with the default timeout. I'm leaning slightly toward connecting, because people are going to use this in a lambda without thinking about it. I think we need more input though.

My sentiment is that we have basically aligned that when in doubt, we should be defaulting to eager connection.

nand4011 (approved):

nand4011:
I guess I'm still recovering from the weekend. I was assuming null == don't eagerly connect.

PR #320, opened 2023-11-07
feat: add updateTtl and itemGetTtl APIs

pratik151192:
Added `updateTtl` and `itemGetTtl` APIs. They're needed for momento lock client https://github.com/momentohq/dev-eco-issue-tracker/issues/548


momento-sdk/src/main/java/momento/sdk/responses/cache/ttl/ItemGetTtlResponse.java
```diff
@@ -0,0 +1,74 @@
...
+     * Gets the retrieved ttl as a Duration.
+     *
+     * @return the ttl.
+     */
+    public Duration ttl() {
+      return ttl;
+    }
+
+    /**
+     * Gets the retrieved ttl as milliseconds since epoch
```
cprice404, line 36: is this actually since epoch? I thought it was remaining TTL millis; that's how it is documented in JS SDK?

pratik151192: Ah yeah you're right good catch. Just typed it along

pratik151192: @cprice404 what do you think about calling this helper method `remainingTtlMillis` and the Duration one as `remainingTtl`

pratik151192: uh thats how it is in JS SDk :)

cprice404 (approved):
lgtm. I think @nand4011 has the most experience with this SDK so I think he should probably take a quick peek too but I'm 👍

nand4011 (approved):

PR #321, opened 2023-11-22
feat: add batch utils with batch get to Java SDK

pratik151192:


momento-sdk/src/main/java/momento/sdk/batchutils/MomentoBatchUtils.java
```diff
@@ -0,0 +1,140 @@
...
+   * @param cacheName The name of the cache.
+   * @param request The batch get request with String keys.
+   * @return BatchGetResponse The batch get response
+   */
+  public BatchGetResponse batchGet(
+      final String cacheName, final BatchGetRequest.StringKeyBatchGetRequest request) {
+
+    if (request.getKeys().size() > this.maxConcurrentRequests) {
+      return maxConcurrentRequestExceededError();
+    }
```
malandis, line 91: As a user I would expect this to iterate over `this.maxConcurrentRequests` batches of keys instead of erroring out.

momento-sdk/src/main/java/momento/sdk/batchutils/MomentoBatchUtils.java
```diff
@@ -0,0 +1,140 @@
...
+        new ArrayList<>();
+
+    for (final byte[] key : request.getKeys()) {
+      CompletableFuture<GetResponse> getResponseFuture = cacheClient.get(cacheName, key);
+      summaries.add(
+          new BatchGetResponse.ByteArrayKeyBatchGetSummary.GetSummary(key, getResponseFuture));
+    }
+
+    return new BatchGetResponse.ByteArrayKeyBatchGetSummary(summaries);
+  }
```
malandis, line 128: This is nearly identical to the above string implementation. Is there no way to abstract away the common parts?

malandis (commented):
Looks good overall; one question about UX and another about DRY code

momento-sdk/src/main/java/momento/sdk/batchutils/MomentoBatchUtils.java
pratik151192: I did start with generics but Javas type erasure kicked in. I'll give it another shot.

momento-sdk/src/main/java/momento/sdk/batchutils/MomentoBatchUtils.java
pratik151192: Didn't completely follow here. This is a mechanism to ensure that users know what they're doing if they give like 1000 keys in the request, which will more often than not overwhelm their application as all the underlying calls are async.

pratik151192: This can be overriden in the builder.

momento-sdk/src/main/java/momento/sdk/batchutils/MomentoBatchUtils.java
pratik151192: Though I can atleast extract out a few lines I agree :)

momento-sdk/src/main/java/momento/sdk/batchutils/MomentoBatchUtils.java
malandis: What I meant was: in multiprocessing libraries where you specify a max job concurrency, usually when you request more jobs than the max concurrency, the remaining jobs are queued and processed once the original jobs begin to complete.

So if I set max concurrent requests to 10 and request 100 keys, I would expect 10 concurrent requests at the start, then as those complete, more requests from the remaining 90 are processed etc until completion.

pratik151192: Ah gotcha. I actually started with a thread pool that would have mimiced the behavior that you mention. But that would have incurred more configuration on the client side. Now that I look at other SDKs, since they are event loop based like Go and JS, they conform to the behavior you call out without any thread pool etc. we could implement a thread pool here and give make the max conn the thread pool size. Thoughts?

malandis: Sounds great 👍

momento-sdk/src/main/java/momento/sdk/batchutils/MomentoBatchUtils.java
```diff
@@ -0,0 +1,188 @@
...
+
+    final List<BatchGetResponse.StringKeyBatchGetSummary.GetSummary> summaries = new ArrayList<>();
+
+    for (final String key : request.getKeys()) {
+      final CompletableFuture<GetResponse> getResponseFuture;
+      try {
+        getResponseFuture =
+            executorService
+                .submit(() -> cacheClient.get(cacheName, key))
+                .get(this.requestTimeoutSeconds, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
pratik151192: This threadpool and the corresponding `get` call is blocking on the _submission_ of the get request, rather than the get request to the server. In theory this will block but not on any network or disk; the `cacheClient.get` call also returns a `Future`, which is returned to the client within the response, and doing a `get()` or `join()` on that will actually block until the server responds back with a result. That is something clients can choose to parallelize, or sequentially process. 

In hindsight, the async nature of our APIs made me choose the decision of wrapping futures within the response, rather than us blocking on them. This threadpool simply makes sure that we _submit_ requests upto `maxConn` in parallel.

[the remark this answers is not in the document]
malandis: So it sounds like the calling thread blocks until the cache completable future is initiated, but not necessarily completed. The client consuming the summaries does so when calling join. But doesn't this mean there will be more thanax concurrent requests in flight?

Ie the code will initiate at most that many requests concurrently but you can have more in flight?

As an alternative I expected the get call as part of the lambda passed to the executor. That way there would be at most max concurrent requests in flight at a time.

I realize this is subtle and may be easier to explain synchronously so let me know :-)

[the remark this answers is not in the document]
pratik151192: > But doesn't this mean there will be more thanax concurrent requests in flight?

Yes, there could be. It would have been much easier to work around this if the SDK APIs were sync. One of the other ways we can do this to be accurate is to use the `ScsDataClient` here directly than using the SDK APIs as the former is sync. 

>  realize this is subtle and may be easier to explain synchronously so let me know 

Yes absolutely, will ping you in Seattle AM time to go over this.

momento-sdk/src/main/java/momento/sdk/batchutils/MomentoBatchUtils.java
```diff
@@ -0,0 +1,229 @@
...
+
+    for (final byte[] key : request.getKeys()) {
+      final CompletableFuture<BatchGetResponse.ByteArrayKeyBatchGetSummary.GetSummary>
+          futureSummary =
+              CompletableFuture.supplyAsync(
+                  () -> {
+                    try {
+                      GetResponse getResponse =
+                          cacheClient
+                              .get(cacheName, key)
```
pratik151192, line 203: This call is an hindrance for me to refactor common logic out as the SDK doesn't implement Java generics. So if I add a generic wrapper for my BatchGetRequest, we do not know here the type.

momento-sdk/src/main/java/momento/sdk/batchutils/MomentoBatchUtils.java
```diff
@@ -0,0 +1,229 @@
...
+    // chain all the futures to generate a new future returned back to the caller
+    return CompletableFuture.allOf(futureSummaries.toArray(new CompletableFuture[0]))
+        .thenApply(
+            v -> {
+              List<BatchGetResponse.StringKeyBatchGetSummary.GetSummary> summaries =
+                  futureSummaries.stream()
+                      .map(CompletableFuture::join)
+                      .collect(Collectors.toList());
+              return new BatchGetResponse.StringKeyBatchGetSummary(summaries);
+            });
```
malandis, line 179: Last request: could you order the keys in the summary identically to the order they are passed in?

malandis (commented):
This is much better! One last feature request regarding key ordering.

momento-sdk/src/main/java/momento/sdk/batchutils/MomentoBatchUtils.java
malandis: Same goes for the bytes summary.

pratik151192: Sure, a linked hash map should do the trick

malandis (approved):
Looks great!

pratik151192:
@malandis Thanks for some really insightful feedback here!

PR #325, opened 2023-11-29
chore: add examples for batch utils and rename auth token to API key

pratik151192:
Add examples for batch utils and rename auth token to API key

### DIFF
--- a/examples/cache/src/main/java/momento/client/example/BatchUtilsExample.java
+++ b/examples/cache/src/main/java/momento/client/example/BatchUtilsExample.java
@@ -1,7 +1,8 @@
 package momento.client.example;
 
 import java.time.Duration;
-import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.List;
 import momento.sdk.CacheClient;
 import momento.sdk.auth.CredentialProvider;
 import momento.sdk.auth.EnvVarCredentialProvider;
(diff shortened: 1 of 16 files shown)


examples/cache/src/main/java/momento/client/example/BatchUtilsExample.java
```diff
@@ -0,0 +1,106 @@
...
+    try (final CacheClient client =
+        CacheClient.create(credentialProvider, Configurations.Laptop.latest(), DEFAULT_ITEM_TTL)) {
+
+      createCache(client, CACHE_NAME);
+      printStartBanner();
+
+      try (final MomentoBatchUtils momentoBatchUtils = MomentoBatchUtils.builder(client).build()) {
+        setupTestData(client, CACHE_NAME);
+        performBatchGet(momentoBatchUtils, CACHE_NAME);
+      }
```
malandis, line 48: At first glance I had a hard time following where the batch builder was instantiated; I was looking for it in the perform batch get method. That said I understand that you've organized like this because it's part of the "setup" code.

Suggestions:
1. Could you add comment in that method with a reference to this block for instantiating it, and also mention the use of the `try` block because the builder is `Closeable`?
2. Make explicit the max number of concurrent requests as a member field and the number of elements to add and fetch. Eg so I can try adding 10 elements and fetching with a concurrency of 5 vs 2.

malandis (commented):
Looks good just one suggestion for improvement

examples/cache/src/main/java/momento/client/example/BatchUtilsExample.java
pratik151192: yeah usually i like the style with `builder` intended on the new line, but our spotless configs change that. Will address these

PR #328, opened 2023-12-02
feat: add client and api for disposable token

rishtigupta:


momento-sdk/src/main/java/momento/sdk/TokenClient.java
```diff
@@ -0,0 +1,217 @@
...
+import momento.sdk.auth.accessControl.DisposableTokenScope;
+import momento.sdk.auth.accessControl.ExpiresIn;
+import momento.sdk.auth.accessControl.TopicSelector;
+import momento.sdk.exceptions.InvalidArgumentException;
+import momento.sdk.exceptions.UnknownException;
+import momento.sdk.responses.GenerateDisposableTokenResponse;
+import momento.token._GenerateDisposableTokenRequest;
+import momento.token._GenerateDisposableTokenResponse;
+
+public class TokenClient {
```
cprice404, line 26: @nand4011 I think this is intended to be an internal class, and I'm trying to remember what our strategy was for making classes like this invisible / less visible to end users in the IDE? do we make them package-protected or something?

cprice404 (commented):
looks good at first glance! so far i just have the one question about the visibility of the internal token client. Will also pull this branch down and try to experiment with it a bit.

momento-sdk/src/main/java/momento/sdk/AuthClient.java
```diff
@@ -0,0 +1,44 @@
...
+import momento.sdk.auth.accessControl.ExpiresIn;
+import momento.sdk.exceptions.InvalidArgumentException;
+import momento.sdk.responses.GenerateDisposableTokenResponse;
+
+public class AuthClient implements IAuthClient {
+  private final TokenClient tokenClient;
+
+  public AuthClient(TokenClient tokenClient) {
+    this.tokenClient = tokenClient;
+  }
```
cprice404, line 14: the examples now show "best practice" for creating a cache client is to use the new `CacheClient.create` factory function. We did this to support eager connections, which probably isn't as relevant for AuthClient, but I think we probably want the best practice code paths for constructing them to look similar, so we should probably add a `create` method here as well. Any thoughts @nand4011 ?

momento-sdk/src/main/java/momento/sdk/AuthClient.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.concurrent.CompletableFuture;
+import momento.sdk.auth.accessControl.DisposableTokenScope;
+import momento.sdk.auth.accessControl.ExpiresIn;
+import momento.sdk.exceptions.InvalidArgumentException;
+import momento.sdk.responses.GenerateDisposableTokenResponse;
+
+public class AuthClient implements IAuthClient {
+  private final TokenClient tokenClient;
+
+  public AuthClient(TokenClient tokenClient) {
```
cprice404, line 12: hrm I think the token client is supposed to be a private/internal implementation detail, so we would want a constructor that allows you to construct an AuthClient without already having a TokenClient?

momento-sdk/src/main/java/momento/sdk/TokenClient.java
cprice404: yeah it looks like the pattern in this repo so far is to prefix them with `Scs` and make them package-protected so they aren't public outside of the source tree.

momento-sdk/src/main/java/momento/sdk/TokenGrpcStubsManager.java
```diff
@@ -0,0 +1,64 @@
...
+import momento.token.TokenGrpc;
+
+/**
+ * Manager responsible for GRPC channels and stubs for the Disposable Token.
+ *
+ * <p>The business layer, will get request stubs from this layer. This keeps the two layers
+ * independent and any future pooling of channels can happen exclusively in the manager without
+ * impacting the API business logic.
+ */
+final class TokenGrpcStubsManager implements Closeable {
```
cprice404, line 22: for consistency with the other internal stubs managers we might want to prefix this with `Scs`.

Alternately since the name `Scs` is kind of outdated we could change the prefix for all of the internal classes to be `Internal` or something.

momento-sdk/src/main/java/momento/sdk/IAuthClient.java
```diff
@@ -0,0 +1,15 @@
+package momento.sdk;
+
+import java.util.concurrent.CompletableFuture;
+import momento.sdk.auth.accessControl.DisposableTokenScope;
+import momento.sdk.auth.accessControl.ExpiresIn;
+import momento.sdk.responses.GenerateDisposableTokenResponse;
+
+public interface IAuthClient extends AutoCloseable {
```
cprice404, line 8: we are using AutoCloseable here, but just `Closeable` on the `CacheClient`. This seems to cause a difference in how error handling works. Not quite sure if one is better than the other, maybe @nand4011 has input.

momento-sdk/src/main/java/momento/sdk/AuthClient.java
```diff
@@ -0,0 +1,44 @@
...
+      ValidationUtils.checkValidDisposableTokenExpiry(expiresIn);
+    } catch (Exception e) {
+      return CompletableFuture.completedFuture(
+          new GenerateDisposableTokenResponse.Error(new InvalidArgumentException(e.getMessage())));
+    }
+    return tokenClient.generateDisposableToken(scope, expiresIn);
+  }
+
+  @Override
+  public void close() throws Exception {
```
cprice404, line 41: re: my comment above about `AutoCloseable` vs `Closeable` - I think actually the difference I am seeing in the error handling is because you put `throws Exception` here. That means someone can't do `try (final AuthClient authClient = ...) { }` without adding a `catch` clause that catches `Exception`. I think you can remove the `throws` here.

momento-sdk/src/main/java/momento/sdk/IAuthClient.java
cprice404: Seems like `AutoCloseable` is more modern/preferred; perhaps we can see if anything breaks if we switch the main `CacheClient` over to that as well.

momento-sdk/src/main/java/momento/sdk/responses/GenerateDisposableTokenResponse.java
```diff
@@ -0,0 +1,40 @@
...
+import momento.sdk.auth.accessControl.ExpiresAt;
+import momento.sdk.exceptions.SdkException;
+import momento.token._GenerateDisposableTokenResponse;
+
+public interface GenerateDisposableTokenResponse {
+
+  class Success implements GenerateDisposableTokenResponse {
+    public final String authToken;
+    public final String endpoint;
+    public final ExpiresAt expiresAt;
```
cprice404, line 15: if you look at the other java response classes, they all have methods like `valueString()` for accessing the response data. Here, you've exposed these as public properties instead.

I think you probably want to make these properties private and then provide accessor methods. @nand4011 may have thoughts though.

cprice404 (commented):
@rishtigupta I think I'm done with a first pass at this. Mostly just stylistic / package structure nits. Overall it feels pretty good!

Would be nice to get examples added once it stabilizes.

momento-sdk/src/intTest/java/momento/sdk/AuthClientTopicTests.java
```diff
@@ -0,0 +1,515 @@
...
+import momento.sdk.auth.accessControl.CacheSelector;
+import momento.sdk.auth.accessControl.DisposableTokenScopes;
+import momento.sdk.auth.accessControl.ExpiresIn;
+import momento.sdk.auth.accessControl.TopicSelector;
+import momento.sdk.exceptions.MomentoErrorCode;
+import momento.sdk.responses.auth.GenerateDisposableTokenResponse;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class AuthClientTopicTests extends BaseTestClass {
```
bruuuuuuuce, line 16: super nit but I think this should be `AuthClientTests`?

momento-sdk/src/main/java/momento/sdk/AuthClient.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Constructs a AuthClient.
+   *
+   * @param credentialProvider Provider for the credentials required to connect to Momento.
+   */
+  public AuthClient(@Nonnull CredentialProvider credentialProvider) {
+    this.tokenClient = new TokenClient(credentialProvider);
+
+    logger.info("Creating Momento Auth Client");
```
bruuuuuuuce, line 25:
```suggestion
    logger.debug("Creating Momento Auth Client");
```

momento-sdk/src/main/java/momento/sdk/AuthClient.java
```diff
@@ -0,0 +1,69 @@
...
+    logger.debug("Cache endpoint: " + credentialProvider.getCacheEndpoint());
+    logger.debug("Control endpoint: " + credentialProvider.getControlEndpoint());
+    logger.debug("Token endpoint: " + credentialProvider.getTokenEndpoint());
+  }
+
+  /**
+   * Constructs a CacheClient.
```
bruuuuuuuce, line 32:
```suggestion
   * Constructs a AuthClient.
```

momento-sdk/src/main/java/momento/sdk/TokenClient.java
```diff
@@ -0,0 +1,217 @@
...
+    return permissions.build();
+  }
+
+  private PermissionsType disposableTokenPermissionToGrpcPermission(
+      DisposableTokenPermission permission) {
+    PermissionsType.Builder result = PermissionsType.newBuilder();
+    if (permission instanceof DisposableToken.CachePermission) {
+      PermissionsType.CachePermissions.Builder cachePerms =
+          PermissionsType.CachePermissions.newBuilder();
+      switch (((DisposableToken.CachePermission) permission).getRole()) {
```
bruuuuuuuce, line 54: I am a bit of a java :noob:, does java enforce exhaustive switch statements? Do we need a default branch here?

momento-sdk/src/main/java/momento/sdk/TokenClient.java
```diff
@@ -0,0 +1,217 @@
...
+            "Unrecognized cache item specification in cache permission: "
+                + gson.toJson(permission));
+      }
+
+      result.setCachePermissions(cachePerms.build());
+
+    } else if (permission instanceof DisposableToken.TopicPermission) {
+      PermissionsType.TopicPermissions.Builder topicPerms =
+          PermissionsType.TopicPermissions.newBuilder();
+      switch (((DisposableToken.TopicPermission) permission).getRole()) {
```
bruuuuuuuce, line 105: ditto on the exhaustive switch statements

bruuuuuuuce (commented):
Awesome work on this, really nice job on the integration tests 🦾 , left a few small nits, biggest concern is about how java deals with switch statements, and if it enforces exhaustive cases or not. Otherwise LGTM

momento-sdk/src/intTest/java/momento/sdk/AuthClientTopicTests.java
rishtigupta: The tests are split into AuthClientCacheTests and AuthClientTopicTests. Hence, the naming!

bruuuuuuuce: ah ok I see its separated by permissions

momento-sdk/src/main/java/momento/sdk/TokenClient.java
rishtigupta: `
But exhaustiveness checks don't end there! While a default branch will always make a switch exhaustive, it isn't required - if the cases cover all possible values, e.g. of an enum, that suffices
`

[ref](https://nipafx.dev/java-switch/#:~:text=But%20exhaustiveness%20checks%20don%27t%20end%20there!%20While%20a%20default%20branch%20will%20always%20make%20a%20switch%20exhaustive%2C%20it%20isn%27t%20required%20%2D%20if%20the%20cases%20cover%20all%20possible%20values%2C%20e.g.%20of%20an%20enum%2C%20that%20suffices%3A)

momento-sdk/src/main/java/momento/sdk/ScsTokenClient.java
```diff
@@ -0,0 +1,217 @@
...
+import momento.sdk.auth.accessControl.DisposableTokenScope;
+import momento.sdk.auth.accessControl.ExpiresIn;
+import momento.sdk.auth.accessControl.TopicSelector;
+import momento.sdk.exceptions.InvalidArgumentException;
+import momento.sdk.exceptions.UnknownException;
+import momento.sdk.responses.auth.GenerateDisposableTokenResponse;
+import momento.token._GenerateDisposableTokenRequest;
+import momento.token._GenerateDisposableTokenResponse;
+
+public class ScsTokenClient {
```
cprice404, line 26: you don't want this class to be `public`, otherwise it can be referenced directly by users and will show up in IDE hints. Check out the other `Scs*` classes for examples.

momento-sdk/src/main/java/momento/sdk/AuthClient.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * Constructs a AuthClient.
+   *
+   * @param credentialProvider Provider for the credentials required to connect to Momento.
+   * @return AuthClient
+   */
+  public static AuthClient create(@Nonnull CredentialProvider credentialProvider) {
+    return create(credentialProvider);
```
cprice404, line 38: this looks like it's just recursing and would cause a stack overflow?

momento-sdk/src/main/java/momento/sdk/IAuthClient.java
```diff
@@ -0,0 +1,16 @@
+package momento.sdk;
+
+import java.io.Closeable;
+import java.util.concurrent.CompletableFuture;
+import momento.sdk.auth.accessControl.DisposableTokenScope;
+import momento.sdk.auth.accessControl.ExpiresIn;
+import momento.sdk.responses.auth.GenerateDisposableTokenResponse;
+
+public interface IAuthClient extends Closeable {
```
cprice404, line 9: @nand4011 she originally had this extending `AutoCloseable`, but changed it to match your existing stuff. I'm actually wondering if we shouldn't change them all to `AutoCloseable` though. Thoughts?

momento-sdk/src/main/java/momento/sdk/responses/auth/GenerateDisposableTokenResponse.java
```diff
@@ -0,0 +1,52 @@
...
+      return authToken;
+    }
+
+    public String endpoint() {
+      return endpoint;
+    }
+
+    public ExpiresAt expiresAt() {
+      return expiresAt;
+    }
```
cprice404, line 44: these seem good to me now, but @nand4011 you should weigh in if you have any thoughts on the consistency of this response class with the existing response classes.

cprice404 (commented):
just a few more things. lmk when those are addressed and I will pull it down and play with it some more.

Also I'd like @nand4011 to review this before it goes in.

momento-sdk/src/main/java/momento/sdk/IAuthClient.java
nand4011: I don't have a problem with that.

momento-sdk/src/main/java/momento/sdk/ScsTokenClient.java
```diff
@@ -0,0 +1,217 @@
...
+          tokenGrpcStubsManager.getStub().generateDisposableToken(request).get();
+      return CompletableFuture.completedFuture(
+          new GenerateDisposableTokenResponse.Success(response));
+    } catch (Exception e) {
+      return CompletableFuture.completedFuture(
+          new GenerateDisposableTokenResponse.Error(new InvalidArgumentException(e.getMessage())));
+    }
+  }
+
+  CompletableFuture<GenerateDisposableTokenResponse> generateDisposableToken(
```
nand4011, line 183: Could this just call the other method?

nand4011 (commented):
Looks good! Thanks for going through all this. This will be the most verbose disposable token implementation until we make an Ada SDK :-)

cprice404 (approved):
pulled this down again and played with it this morning. looks good to me now!

celebrate

PR #332, opened 2023-12-04
feat: add topic support

rishtigupta:
### PR Description:
- Add topic client with publish and send apis


momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -0,0 +1,207 @@
...
+    } catch (Exception e) {
+      future.completeExceptionally(
+          new TopicSubscribeResponse.Error(CacheServiceExceptionMapper.convert(e)));
+    }
+    return future;
+  }
+
+  private void handleSubscriptionError(Throwable t) {
+    // Handle the subscription error
+    logger.info("error " + cacheName + " " + topicName + " " + t.getMessage());
```
bruuuuuuuce, line 91:
```suggestion
    logger.trace("error " + cacheName + " " + topicName + " " + t.getMessage());
```

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -0,0 +1,207 @@
...
+    this.options.onError(t);
+  }
+
+  private void handleSubscriptionCompleted() {
+    // Handle the subscription completion
+    logger.info("completed " + cacheName + " " + topicName);
```
bruuuuuuuce, line 97:
```suggestion
    logger.trace("completed " + cacheName + " " + topicName);
```

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -0,0 +1,207 @@
...
+    this.options.onCompleted();
+  }
+
+  private void handleSubscriptionItem(_SubscriptionItem item) {
+    // Handle subscription item
+    if (!subscribed) {
```
bruuuuuuuce, line 103: its a bit strange to me to have both a `firstMessage` boolean, as well as a `subscribed` boolean. I would try and consolidate the two. At first glance, I would probably try moving this logic checking for the heartbeat above where we complete the original future. If the first message is not a heartbeat, we should complete the future exceptionally

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -0,0 +1,207 @@
...
+        break;
+      default:
+        handleSubscriptionUnknown();
+        break;
+    }
+  }
+
+  private void handleSubscriptionDiscontinuity(_SubscriptionItem discontinuityItem) {
+    // Handle subscription discontinuity
+    logger.info(
```
bruuuuuuuce, line 136:
```suggestion
    logger.debug(
```

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -0,0 +1,207 @@
...
+        cacheName,
+        topicName,
+        discontinuityItem.getDiscontinuity().getLastTopicSequence(),
+        discontinuityItem.getDiscontinuity().getNewTopicSequence());
+  }
+
+  private void handleSubscriptionHeartbeat() {
+    // Handle subscription heartbeat
+    logger.info("heartbeat " + " " + cacheName + " " + topicName);
```
bruuuuuuuce, line 145:
```suggestion
    logger.debug("heartbeat " + " " + cacheName + " " + topicName);
```

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  private void handleSubscriptionUnknown() {
+    // Handle subscription unknown
+    logger.info("unknown " + cacheName + " " + topicName);
```
bruuuuuuuce, line 150:
```suggestion
    logger.warn("unknown " + cacheName + " " + topicName);
```

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  private TopicMessage.Binary handleSubscriptionBinaryMessage(byte[] binary, String publisherId) {
+    _TopicValue topicValue =
+        _TopicValue.newBuilder().setBinary(ByteString.copyFrom(binary)).build();
+    // Handle subscription binary message
+    return new TopicMessage.Binary(topicValue, publisherId.isEmpty() ? null : publisherId);
+  }
+
+  private void handleSubscriptionUnknownMessage() {
```
bruuuuuuuce, line 189: i think this a duplicate of https://github.com/momentohq/client-sdk-java/pull/332/files#diff-5d034b9e3e73ef3355324703eceb4081fa0db81bf278c7ce03e6b323d3a4f0a3R148?

momento-sdk/src/main/java/momento/sdk/TopicClient.java
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Constructs a TopicClient.
+   *
+   * @param credentialProvider Provider for the credentials required to connect to Momento.
+   * @param configuration Configuration object containing all tunable client settings.
+   */
+  public TopicClient(
+      @Nonnull CredentialProvider credentialProvider, @Nonnull Configuration configuration) {
+    this.scsTopicClient = new ScsTopicClient(credentialProvider, configuration);
+    logger.info("Creating Momento Topic Client");
```
bruuuuuuuce, line 28:
```suggestion
    logger.debug("Creating Momento Topic Client");
```

momento-sdk/src/main/java/momento/sdk/TopicClient.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.concurrent.CompletableFuture;
+import javax.annotation.Nonnull;
+import momento.sdk.auth.CredentialProvider;
+import momento.sdk.config.Configuration;
+import momento.sdk.responses.topic.TopicPublishResponse;
+import momento.sdk.responses.topic.TopicSubscribeResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Client to perform operations against the Momento Cache Service */
```
bruuuuuuuce, line 13:
```suggestion
/** Client to perform operations against the Momento Topic Service */
```

momento-sdk/src/main/java/momento/sdk/ISubscribeCallOptions.java
```diff
@@ -0,0 +1,11 @@
+package momento.sdk;
+
+import momento.sdk.responses.topic.TopicMessage;
+
+public interface ISubscribeCallOptions {
```
bruuuuuuuce, line 5: are we able to add the `resumeAtTopicSequenceNumber` to this interface? I am not quite sure how java interfaces work, but we need to add that to the `subscribe` method as well

momento-sdk/src/intTest/java/momento/sdk/TopicClientTest.java
```diff
@@ -0,0 +1,161 @@
...
+    TopicSubscribeResponse response = topicClient.subscribe(cacheName, topicName, options).join();
+    logger.info(response.toString());
+    assertThat(response).isInstanceOf(TopicSubscribeResponse.Subscription.class);
+
+    TopicPublishResponse publishResponse = topicClient.publish(cacheName, topicName, value).join();
+    logger.info(publishResponse.toString());
+    assertThat(publishResponse).isInstanceOf(TopicPublishResponse.Success.class);
+
+    Thread.sleep(1000);
+    ((TopicSubscribeResponse.Subscription) response).unsubscribe();
```
bruuuuuuuce, line 159: ditto

bruuuuuuuce (commented):
nice work 🦾 , left some feedback and nit, looks like you are on the right track :w00t:

momento-sdk/src/main/java/momento/sdk/ISubscribeCallOptions.java
rishtigupta: We have it in subscription state. Javascript one does not pass the param in the subscribe method. I was confused on this as well. @cprice404 any suggestions?

momento-sdk/src/intTest/java/momento/sdk/TopicClientTest.java
```diff
@@ -0,0 +1,161 @@
...
```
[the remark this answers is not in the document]
nand4011: +1. You could create an onItem callback that performs assertions and maybe decrements a CountDownLatch so you can pause until the callback triggers.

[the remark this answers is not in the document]
rishtigupta: yes, done!

momento-sdk/src/main/java/momento/sdk/responses/topic/SubscriptionState.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public int getResumeAtTopicSequenceNumber() {
+    return (lastTopicSequenceNumber != null ? lastTopicSequenceNumber : -1) + 1;
+  }
+
+  public void setSubscribed() {
+    this.isSubscribed = true;
+  }
+
+  public void setUnsubscribed() {
```
nand4011, line 21: When would I use this vs `unsubscribe()`? You might want to add some javadocs here since there are similarly named functions that do different things.

momento-sdk/src/main/java/momento/sdk/ScsTopicGrpcStubsManager.java
```diff
@@ -0,0 +1,66 @@
...
+   *
+   * <p>Each stub is deliberately decorated with Deadline. Deadlines work differently than timeouts.
+   * When a deadline is set on a stub, it simply means that once the stub is created it must be used
+   * before the deadline expires. Hence, the stub returned from here should never be cached and the
+   * safest behavior is for clients to request a new stub each time.
+   *
+   * <p><a href="https://github.com/grpc/grpc-java/issues/1495">more information</a>
+   */
+  PubsubGrpc.PubsubStub getStub() {
+    return stub.withDeadlineAfter(deadline.getSeconds(), TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
rishtigupta: Yes, the implementation that I had before kept sending heartbeat messages due to which test was timing out after 15 sec.

momento-sdk/src/main/java/momento/sdk/responses/topic/SubscriptionState.java
rishtigupta: setUnsubscribed sets the subscribed object to false when unsubscribe function is called. Added java docs which should make it more clear.

momento-sdk/src/main/java/momento/sdk/responses/topic/SubscriptionState.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  public void setUnsubscribeFn(Runnable unsubscribeFn) {
+    this.unsubscribeFn = unsubscribeFn;
+  }
+
+  /** Unsubscribes from the topic, executing the unsubscribe function. */
+  public void unsubscribe() {
+    if (isSubscribed) {
+      unsubscribeFn.run();
+      setUnsubscribed();
```
bruuuuuuuce, line 57: we can just have `this.isSubscribed = false; and remove the `setUnsubscribed` method?

momento-sdk/src/intTest/java/momento/sdk/TopicClientTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+    TopicSubscribeResponse response = topicClient.subscribe(cacheName, topicName, options).join();
+    logger.info(response.toString());
+    assertThat(response).isInstanceOf(TopicSubscribeResponse.Subscription.class);
+
+    TopicPublishResponse publishResponse = topicClient.publish(cacheName, topicName, value).join();
+    logger.info(publishResponse.toString());
+    assertThat(publishResponse).isInstanceOf(TopicPublishResponse.Success.class);
+
+    TimeUnit.MILLISECONDS.sleep(1000);
```
bruuuuuuuce, line 178: thread.sleep inside of a test is usually a code smell. I liked nates suggestion of using countdown latches to await. Here is an example of using a similar mechanism in kotlin

https://github.com/momentohq/cache-admin/blob/f237d878eeb0108633bff79aeaba4a34b8837491/kotlin/shared/control-plane/src/test/kotlin/momento/controlplane/ControlPlaneActionsTest.kt#L353-L400

More docs on this using java

https://onraz.github.io/java/2014/08/05/testing-asynchronous-code-in-java-with-countdownlatch.html

momento-sdk/src/main/java/momento/sdk/ScsTopicClient.java
```diff
@@ -0,0 +1,132 @@
...
+  private final ScsTopicGrpcStubsManager topicGrpcStubsManager;
+  private final CredentialProvider credentialProvider;
+
+  public ScsTopicClient(
+      @Nonnull CredentialProvider credentialProvider, @Nonnull Configuration configuration) {
+    this.credentialProvider = credentialProvider;
+    this.topicGrpcStubsManager = new ScsTopicGrpcStubsManager(credentialProvider, configuration);
+  }
+
+  public CompletableFuture<TopicPublishResponse> publish(
```
bruuuuuuuce, line 27: I think we should add javadocs to these methods as well since they will be the ones customers will be directly calling

momento-sdk/src/main/java/momento/sdk/ScsTopicClient.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  private CompletableFuture<TopicSubscribeResponse> sendSubscribe(
+      String cacheName, String topicName, ISubscribeCallOptions options) {
+    SubscriptionWrapper subscriptionWrapper;
+    SubscriptionState subState = new SubscriptionState();
+    subscriptionWrapper =
+        new SubscriptionWrapper(topicGrpcStubsManager, cacheName, topicName, subState, options);
+    final CompletableFuture<Void> subscribeFuture = subscriptionWrapper.subscribe();
+    return subscribeFuture.handle(
```
bruuuuuuuce, line 117: 🦾

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -0,0 +1,191 @@
...
+
+    subscription =
+        new StreamObserver<_SubscriptionItem>() {
+          boolean firstMessage = true;
+
+          @Override
+          public void onNext(_SubscriptionItem item) {
+            if (firstMessage) {
+              if (item.getKindCase() != _SubscriptionItem.KindCase.HEARTBEAT) {
+                throw new InternalServerException(
```
bruuuuuuuce, line 51: should this be `future.completeExceptionally(new InternalServerError(...))`?

momento-sdk/src/main/java/momento/sdk/TopicClient.java
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Constructs a TopicClient.
+   *
+   * @param credentialProvider Provider for the credentials required to connect to Momento.
+   * @param configuration Configuration object containing all tunable client settings.
+   * @return TopicClient
+   */
+  public static TopicClient create(
+      @Nonnull CredentialProvider credentialProvider, @Nonnull Configuration configuration) {
```
bruuuuuuuce, line 41: In our other sdks, we distinguish between `Configuration`, which is a cache configuration, and `TopicConfiguration` which are configurations specific to the topic client

momento-sdk/src/main/java/momento/sdk/responses/topic/SubscriptionState.java
```diff
@@ -0,0 +1,60 @@
+package momento.sdk.responses.topic;
+
+/** Represents the state of a subscription to a topic. */
+public class SubscriptionState {
+
+  private Runnable unsubscribeFn;
+  private Integer lastTopicSequenceNumber;
+  private boolean isSubscribed;
```
bruuuuuuuce, line 8: I actually don't think that this boolean is needed. The only place it is used is

```
    if (isSubscribed) {
      unsubscribeFn.run();
      setUnsubscribed();
```

Instead we could do something like

```
```
  unsubscribeFn.run();
  this.unsubscribeFn = () -> {}
```

This would still ensure that each `unsubscribeFn` was only called once, but I think would meaningfully simplify this class

File: momento-sdk/src/main/java/momento/sdk/responses/topic/SubscriptionState.java (hunk 15 above)
Comment (rishtigupta) [reply momento-sdk/src/main/java/momento/sdk/responses/topic/SubscriptionState.java:57]: Wanted to be consistent with javascript sdk. But if you think, that's a better way, then I can definitely change it

File: momento-sdk/src/intTest/java/momento/sdk/TopicClientTest.java (hunk 16 above)
Comment (rishtigupta) [reply momento-sdk/src/intTest/java/momento/sdk/TopicClientTest.java:178]: Good catch!

File: momento-sdk/src/main/java/momento/sdk/TopicClient.java (hunk 20 above)
Comment (rishtigupta) [reply momento-sdk/src/main/java/momento/sdk/TopicClient.java:41]: I will add TopicConfiguration to our configurations and use that for the topic clients. Thanks!

File: momento-sdk/src/main/java/momento/sdk/ISubscribeCallOptions.java [hunk 22]
```diff
@@ -0,0 +1,23 @@
...
+import momento.sdk.responses.topic.TopicMessage;
+
+/** Represents options for a topic subscription callback. */
+public interface ISubscribeCallOptions {
+  /**
+   * Called when a new message is received on the subscribed topic.
+   *
+   * @param message The received topic message.
+   */
+  void onItem(TopicMessage message);
```
cprice404, line 12: you are doing callback-style here. In some SDKs we do callback-style, and in some we do iterator style with a blocking iterator method like `nextMessage()`. I believe golang and .NET do that iterator style.

This usually boils down to how the gRPC interface for streams is presented to us by the grpc lib. I assume in Java it's callback-style?

@nand4011 I believe that you were the one that did the iterator-style implementation for .NET, right? can you weigh in on whether that pattern would be possible/desirable in Java, based on what you've seen of the grpc lib's support for streaming?

I'm not opposed to callback-style, but just want to make sure we have investigated the options and made a deliberate, informed choice.

momento-sdk/src/main/java/momento/sdk/responses/topic/SubscriptionState.java
rishtigupta: The boolean is also used to mark the subscription to true when subscribing using the grpc client

momento-sdk/src/main/java/momento/sdk/TopicClient.java
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Publish a message to a topic with provided topic name in a cache with provided cache name.
+   *
+   * @param cacheName The name of the cache where topic resides.
+   * @param topicName The name of the topic.
+   * @param message The message to be published.
+   * @return A future containing the result of the topic publish: {@link
+   *     TopicPublishResponse.Success} or {@link TopicPublishResponse.Error}.
+   */
+  public CompletableFuture<TopicPublishResponse> publish(
```
cprice404, line 66: when playing with error handling on these I noticed that if I specify a cache that does not exist, it seems that we throw the NOT_FOUND exception (as opposed to returning an `Error` response).

I am okay with this as long as:
(a) this is how we are handling it elsewhere in the Java SDK (e.g. cache set/get), and
(b) we are returning the Error response type for other errors? otherwise it doesn't make any sense to even have the error response type if we are just always going to throw.

can you verify this?

momento-sdk/src/main/java/momento/sdk/TopicClient.java
```diff
@@ -0,0 +1,104 @@
...
+   * Subscribe to a topic with provided topic name in a cache with provided cache name.
+   *
+   * @param cacheName The name of the cache where topic resides.
+   * @param topicName The name of the topic.
+   * @param options The options for handling the subscription callbacks. See {@link
+   *     ISubscribeCallOptions}.
+   * @return A future containing the result of the topic subscribe: {@link
+   *     TopicSubscribeResponse.Subscription} or {@link TopicSubscribeResponse.Error}.
+   */
+  public CompletableFuture<TopicSubscribeResponse> subscribe(
```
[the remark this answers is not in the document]
pratik151192: 100% agreed on `subscribeResponse.orElseThrow(() -> new RuntimeException("Subscription failed: " + subscribeResponse));` its more idiomatic and was released in Java 8; so its also quite well accepted within the community. The only argument that folks might have is that we do require casting for our cache methods, so users of both topic and cache clients would have a dual experience. Although with caches we have established a sealed class pattern with different possibilities that don't apply here, so we can live with that. I'd still +1 on the Optional as it looks way neater.

momento-sdk/src/main/java/momento/sdk/TopicClient.java
cprice404: I modified my program to run a bunch of threads that all do Publishes in a loop, to intentionally exceed the throttling limit, and AFAICT the first time I hit the throttling limit an exception is thrown (rather than an Error response being returned).

This seems inconsistent with other SDKs as far as I know? Consistency is my top priority so if I'm wrong about what happens in other SDKs we can discuss.

But unless i'm doing something silly, it seems like the `Error` types in this PR may not be getting used at all yet.

momento-sdk/src/main/java/momento/sdk/TopicClient.java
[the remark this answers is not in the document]
bruuuuuuuce: I am a complete java :noob: so if what I am saying doesnt make sense can ignore it, but that looks like to me that `.orElseThrow` violates our pattern set of explicitly _not_ throwing exceptions. To me its weird that none of our methods throw exceptions except this one? Or maybe `.orElseThrow` is doing something that I am not expecting?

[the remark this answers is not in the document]
bruuuuuuuce: To clarify, why would we throw an exception rather that returning the error response?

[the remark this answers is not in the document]
pratik151192: Yeah fair that's what I said that it's a valid argument that we are going to deviate from some of the established patterns. I think we could still use optional and use `orElse(return Error)` instead of `orElseThrow` to your concern Matt.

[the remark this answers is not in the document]
bruuuuuuuce: Ok cool that approach seems more consistent to me, I like it 👍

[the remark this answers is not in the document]
cprice404: > I am a complete java :noob: so if what I am saying doesnt make sense can ignore it, but that looks like to me that .orElseThrow violates our pattern set of explicitly not throwing exceptions. 

in this case the user is explicitly telling us they want to throw so I think it would be okay. but we could totally support both `orElse` and `orElseThrow`, they're not exclusive.

> The only argument that folks might have is that we do require casting for our cache methods, so users of both topic and cache clients would have a dual experience. 

I don't think this is necessarily true? If we have any "best practice"/example code for our existing APIs that requires explicit casting, then I would want to provide alternatives for that ASAP. In the switch statements I think it's fine, when the language itself is doing like, "smart casting", but if there is anywhere that a user can't use our SDK with out doing an explicit cast via `()` then I would like to dig into that so please let me know.

momento-sdk/src/main/java/momento/sdk/ISubscribeCallOptions.java
```diff
@@ -0,0 +1,23 @@
+package momento.sdk;
+
+import momento.sdk.responses.topic.TopicMessage;
+
+/** Represents options for a topic subscription callback. */
+public interface ISubscribeCallOptions {
```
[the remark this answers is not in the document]
nand4011: The 'options' naming idiom seems pretty specific to Javascript. I like the idea of something more descriptive like `ISubscriptionCallbacks`, unless we will need to add non-callback fields to it in the future that can't go in the method signature.

momento-sdk/src/main/java/momento/sdk/ISubscribeCallOptions.java
nand4011: It looks like we can supply a `ClientCallStreamObserver` to the request that lets you disable automatic message requesting and provides a request method so you can force it into an iterator style pattern.

cprice404: @nand4011 cool, thanks. since you are among our most experienced Java engineers and you also worked on the iterator-style pattern in the .NET SDK, do you have thoughts on whether it's worth it?

Also do we know what thread the callbacks get called on if we stick with the callback pattern?

momento-sdk/src/main/java/momento/sdk/responses/topic/TopicSubscribeResponse.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * Returns the Subscription if the response is a Subscription, otherwise returns the provided
+   * default value.
+   *
+   * @param other The default value to be returned if the response is not a Subscription.
+   * @return The Subscription if the response is a Subscription, otherwise the default value.
+   */
+  default Subscription orElse(Subscription other) {
```
cprice404, line 53: In the thread I know we were suggesting implementing `orElse` but now that I'm looking at this I realize that it's extremely unlikely anyone would ever use this particular signature because they would need to have an existing `Subscription` object in scope to fall back to, and I can't really imagine that ever happening. I think we were advocating for the `Supplier` version of the signature so that the caller could pass in a lambda function that would do some kind of error handling and/or throw an exception. That's what `orElseGet` does, which you implemented below. So I think we should probably just remove this method for now.  @nand4011 ?

momento-sdk/src/main/java/momento/sdk/responses/topic/TopicSubscribeResponse.java
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Returns the Subscription if the response is a Subscription, otherwise throws an exception
+   * produced by the specified Supplier.
+   *
+   * @param <X> The type of the exception to be thrown.
+   * @param exceptionSupplier The Supplier that produces the exception to be thrown.
+   * @return The Subscription if the response is a Subscription.
+   * @throws X if the response is not a Subscription.
+   */
+  default <X extends Throwable> Subscription orElseThrow(Error exceptionSupplier) throws X {
```
cprice404, line 78: this signature should take a `Supplier<? extends X>`, rather than just an Error.

momento-sdk/src/intTest/java/momento/sdk/TopicClientTest.java
```diff
@@ -0,0 +1,192 @@
...
+    String value = "test-value";
+
+    TopicSubscribeResponse response = topicClient.subscribe(cacheName, topicName, options).join();
+    assertThat(response).isInstanceOf(TopicSubscribeResponse.Subscription.class);
+
+    TopicPublishResponse publishResponse = topicClient.publish(cacheName, topicName, value).join();
+    assertThat(publishResponse).isInstanceOf(TopicPublishResponse.Success.class);
+
+    latch.countDown();
+    latch.await();
```
cprice404, line 180: I'm worried that this isn't doing what you think it is doing.

AFAICT the latch is getting initialized once, as a class-level variable, with its count set to 2.

Then there are several places where you call `countDown`, and they are spread out over a few places. One happens in the callback and the others happen in various tests. I think it's probably getting called 4 times. After the first 2, the calls to `countDown` and `await` are no-ops.

Using a countdown latch for a test like these is a very reasonable thing to do but you will want to create a new latch for each test, so that the scope of the latch variable is limited to the individual test.

I think you can just change your hard-coded `options` variable to be a method that returns a new options, instead, like:

`ISusbcriptionCallbacks callbacks(CountDownLatch latch) { return new ISubscriptionCallbacks(...); }`

For each test, the latch you pass in should only need 1 countdown per message that you are sending, so probably for both of these tests they would just pass in `new CountDownLatch(1)`. Then you will do the `countDown` in the `onItem` callback, and you will only need an `await` in the test itself (not another `countDown`).
And then you

momento-sdk/src/main/java/momento/sdk/ScsTopicClient.java
```diff
@@ -0,0 +1,134 @@
...
+
+                @Override
+                public void onNext(Object value) {
+                  // Do nothing
+                }
+
+                @Override
+                public void onError(Throwable t) {
+                  future.completeExceptionally(
+                      new TopicPublishResponse.Error(CacheServiceExceptionMapper.convert(t)));
```
cprice404, line 95: I think that if you want the end user to get an instance of `Error` without it being thrown as an exception, you probably don't want to call `completeExceptionally` here? I can test it again with that program that I pasted that intentionally causes throttles. or we can test with a bad cache name - I can't remember what is supposed to happen in other SDKs if you do `publish` with a cache name that doesn't exist. If that returns an `Error` response type in the other SDKs then maybe we should add a test for that case in your code, to verify how these completes are working.

momento-sdk/src/main/java/momento/sdk/config/TopicConfigurations.java
```diff
@@ -0,0 +1,120 @@
...
+      return Laptop.v1();
+    }
+
+    /**
+     * Provides v1 configuration for a laptop environment. This configuration is guaranteed not to
+     * change in future releases of the Momento Java SDK.
+     *
+     * @return the v1 laptop configuration
+     */
+    public static TopicConfiguration v1() {
```
cprice404, line 40: let's not include the v1s yet. We can just provide `latest` and add a v1 once some users have gotten some mileage on it.

momento-sdk/src/main/java/momento/sdk/responses/topic/SubscriptionState.java
```diff
@@ -0,0 +1,51 @@
+package momento.sdk.responses.topic;
+
+/** Represents the state of a subscription to a topic. */
+public class SubscriptionState {
```
cprice404, line 4: I don't think this class needs to be public - we don't expect an end user to ever see it or touch it, right?

momento-sdk/src/main/java/momento/sdk/responses/topic/TopicMessage.java
```diff
@@ -0,0 +1,66 @@
...
+      return value;
+    }
+
+    public String getTokenId() {
+      return tokenId;
+    }
+
+    @Override
+    public String toString() {
+      return "Binary{" + "value=" + Arrays.toString(value) + ", tokenId='" + tokenId + '\'' + '}';
```
cprice404, line 51: that `Arrays.toString` may not produce a useful representation of the data, and we may need to truncate these so that they don't blow out the console/logs for large items, but we can tweak that later once we have played with it more to see how it feels.

cprice404 (commented):
looking good, left a few more suggestions / comments

momento-sdk/src/main/java/momento/sdk/responses/topic/SubscriptionState.java
cprice404: @nand4011 - I was on a zoom with @rishtigupta and she pointed out that we can't simply remove the 'public' modifier for this, even though it's not supposed to be visible to end users, because it gets used by other classes in our code that do not live in this package.

Do you know of any java paradigms I'm overlooking here that would allow us to prevent this class from being visible in the SDK while still allowing us to reference it from both the Response.Subscription class and our internal Client classes?

nand4011: Not off hand. It's similar to all of the ScsClient and stub manager classes being public. I haven't used modules much, but we may be able to define the public API using one of those.

cprice404: the ScsClient and StubManager classes aren't public though :)

nand4011: Ha, fair enough. I was thinking about them being at the top level instead of in the internal package.

PR #337, opened 2023-12-13
feat: add error handling topics

rishtigupta:
## PR Description:
- Use ClientCallStreamObserver instead of StreamObserver to observe the subscription response
- Use the cancel api from the grpc's ClientCallStreamObserver to unsubscribe as opposed to killing the TCP connection by closing in the GRPC connection on unsubscribe
- Add subscription retries when the server is UNAVAILABLE (due to connection loss/other reasons)


momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -19,42 +21,54 @@
...
+  public SubscriptionWrapper(ScsTopicGrpcStubsManager grpcManager, SendSubscribeOptions options) {
     this.grpcManager = grpcManager;
-    this.cacheName = cacheName;
-    this.topicName = topicName;
-    this.subscriptionState = subscriptionState;
     this.options = options;
   }
 
-  public CompletableFuture<Void> subscribe() {
+  public CompletableFuture<Void> subscribeWithRetry() {
```
cprice404, line 34: this should be able to be fully `private` i believe

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -19,42 +21,54 @@
...
     CompletableFuture<Void> future = new CompletableFuture<>();
+    subscribeWithRetryInternal(future);
+    return future;
+  }
 
+  public void subscribeWithRetryInternal(CompletableFuture<Void> future) {
```
cprice404, line 40: `private`

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -19,42 +21,54 @@
...
+          public void setOnReadyHandler(Runnable onReadyHandler) {}
+
+          @Override
+          public void request(int count) {}
+
+          @Override
+          public void setMessageCompression(boolean enable) {}
+
+          @Override
+          public void disableAutoInboundFlowControl() {}
```
cprice404, line 60: can you push these overrides up into your `CancelableClientCallStreamObserver` class, so that consumers of it don't have to worry about them?

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -70,6 +84,16 @@ public void onError(Throwable t) {
               firstMessage = false;
               future.completeExceptionally(t);
 
+              // Retry logic for UNAVAILABLE errors
+              if (t instanceof io.grpc.StatusRuntimeException) {
```
cprice404, line 88: let's make sure we have a `logger.debug` statement in each of the branches of all of the conditions here, may be important for debugging.

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -70,6 +84,16 @@ public void onError(Throwable t) {
...
+                io.grpc.StatusRuntimeException statusRuntimeException =
+                    (io.grpc.StatusRuntimeException) t;
+                if (statusRuntimeException.getStatus().getCode() == Status.Code.UNAVAILABLE) {
+                  logger.info("Retrying subscription after a delay...");
```
cprice404, line 92: i think it's probably reasonable to log this at `info`, but if you are going to do that, you need to make sure log something else earlier that shows that an error occurred or whatever, so that users won't just see this log message and wonder "why?"

momento-sdk/src/main/java/momento/sdk/config/TopicConfigurations.java
```diff
@@ -29,7 +29,7 @@ private Laptop(TransportStrategy transportStrategy, Logger logger) {
      */
     public static TopicConfiguration latest() {
       final TransportStrategy transportStrategy =
-          new StaticTransportStrategy(new GrpcConfiguration(Duration.ofMillis(15000)));
+          new StaticTransportStrategy(new GrpcConfiguration(Duration.ofMillis(20_000)));
```
cprice404, line 32: we can probably change this back to 15

momento-sdk/src/main/java/momento/sdk/SendSubscribeOptions.java
```diff
@@ -0,0 +1,90 @@
+package momento.sdk;
+
+import momento.sdk.responses.topic.SubscriptionState;
+import momento.sdk.responses.topic.TopicMessage;
+import momento.sdk.responses.topic.TopicSubscribeResponse;
+
+public class SendSubscribeOptions implements ISubscriptionCallbacks {
```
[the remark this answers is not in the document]
rishtigupta: Correct

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
rishtigupta: private within the package since it is called by the ScsTopicClient class (in the same package though)

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -84,17 +67,24 @@ public void onError(Throwable t) {
               firstMessage = false;
               future.completeExceptionally(t);
 
-              // Retry logic for UNAVAILABLE errors
+              logger.debug("First message failed, retrying subscription...");
```
cprice404, line 70: okay now that I read this message I think this method may not be quite right yet.

If you look at the JS SDK, you will see that the `firstMessage` check is there to try to *prevent* us from retrying the subscription. The use case is for cases like when the cache name is invalid; we don't want to keep retrying forever if the cache name is invalid.

so this code:
https://github.com/momentohq/client-sdk-javascript/blob/dd70f5a97c31b60a5314e922f5759de909bbd1dc/packages/core/src/internal/clients/pubsub/AbstractPubsubClient.ts#L179-L188

is checking to see if we are on the first message, and if we are, then we just allow the failure to propagate to the user. only if we are NOT on the first message do we retry the subscription.

What happens in your code if the cache name is wrong?

(Also if you are going to log that an error happened, you should always include the error object itself in the log message.)

cprice404 (commented):
I think this probably LGTM now pending some investigation/tweaking around that "first message" thing

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
rishtigupta: If the cache nae is wrong it would not retry the subscription. It would retry only when the statusCode is UNAVAILABLE. So, yes I think the log message is misleading but the logic still stands correct. Wdyt?

cprice404: what are you using the `firstMessage` variable for, then?

in the JS SDK the whole reason that that variable exists is to make sure that we do NOT retry connection on firstMessage, and in your code, it reads as though we *only* retry on firstMessage, so something is off I think :)

rishtigupta: After the connection is lost and regained, it enters the "if" block indicating that the code is treating the message just after the connection is back online as the `firstMessage`?

cprice404: what value is `firstMessage` providing you? What is the use case where the behavior would be incorrect if you did not have it?

You currently have the `instanceof StatusRuntimeException` code in two places, one for each branch of your `if (firstMessage)` code, and I don't understand why.

cprice404: in other words, the logic of your if/else branches here looks pretty much identical to me except for the logging.

cprice404: is it just that one code path delays first and the other doesn't?

rishtigupta: Fixed based on offline discussion. Thanks!

cprice404 (approved):
I added a commit that adds an int test for the "cache not found" case just to be safe.

I think this is a good checkpoint, good improvement over current `HEAD`, let's merge it and tackle remaining work in a few follow-up PRs

cprice404 (approved):
I added a commit that adds an int test for the "cache not found" case just to be safe.

I think this is a good checkpoint, good improvement over current HEAD, let's merge it and tackle remaining work in a few follow-up PRs

PR #351, opened 2024-03-05
feat: add grpc config options and turn off keepalive for Lambda config

anitarua:
Adds message size to support messages up to 5mb by default (though I was only able to find the max receive message size option through the [NettyChannelBuilder](https://grpc.github.io/grpc-java/javadoc/io/grpc/netty/NettyChannelBuilder.html#maxInboundMessageSize(int)) interface).

Adds keepalive settings to the configs as well. Disables keepalive for the cache control client and the Lambda prebuilt config. Sets different keepalive values for the topics prebuilt configs based on the values that were hardcoded into the topic grpc manager.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/ScsControlGrpcStubsManager.java
+++ b/momento-sdk/src/main/java/momento/sdk/ScsControlGrpcStubsManager.java
@@ -37,7 +37,7 @@ final class ScsControlGrpcStubsManager implements AutoCloseable {
 
   private static ManagedChannel setupConnection(
       CredentialProvider credentialProvider, Configuration configuration) {
-    final NettyChannelBuilder channelBuilder =
+    NettyChannelBuilder channelBuilder =
         NettyChannelBuilder.forAddress(credentialProvider.getControlEndpoint(), 443);
 
     // Override grpc config to disable keepalive for control clients
@@ -45,7 +45,7 @@ final class ScsControlGrpcStubsManager implements AutoCloseable {
         configuration.getTransportStrategy().getGrpcConfiguration().withKeepAliveDisabled();
 
     // set additional channel options (message size, keepalive, auth, etc)
-    GrpcChannelOptions.GrpcOptionsFromGrpcConfig(controlConfig, channelBuilder);
+    channelBuilder = GrpcChannelOptions.GrpcOptionsFromGrpcConfig(controlConfig, channelBuilder);
 
     final List<ClientInterceptor> clientInterceptors = new ArrayList<>();
     clientInterceptors.add(new UserHeaderInterceptor(credentialProvider.getAuthToken()));
--- a/momento-sdk/src/main/java/momento/sdk/ScsControlGrpcStubsManager.java
+++ b/momento-sdk/src/main/java/momento/sdk/ScsControlGrpcStubsManager.java
@@ -10,6 +10,9 @@ import java.util.List;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
 import momento.sdk.auth.CredentialProvider;
+import momento.sdk.config.Configuration;
+import momento.sdk.config.transport.GrpcConfiguration;
+import momento.sdk.internal.GrpcChannelOptions;
 
 /**
  * Manager responsible for GRPC channels and stubs for the Control Plane.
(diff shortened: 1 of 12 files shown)


momento-sdk/src/main/java/momento/sdk/ScsTopicGrpcStubsManager.java
```diff
@@ -28,20 +28,19 @@ final class ScsTopicGrpcStubsManager implements Closeable {
...
+  private static ManagedChannel setupConnection(
+      CredentialProvider credentialProvider, TopicConfiguration configuration) {
     final NettyChannelBuilder channelBuilder =
         NettyChannelBuilder.forAddress(credentialProvider.getCacheEndpoint(), 443);
-    channelBuilder.useTransportSecurity();
-    channelBuilder.disableRetry();
 
-    channelBuilder.keepAliveTime(10, TimeUnit.SECONDS);
-    channelBuilder.keepAliveTimeout(5, TimeUnit.SECONDS);
-    channelBuilder.keepAliveWithoutCalls(true);
```
anitarua, line 44: propagated these values to the prebuilt topics configs

momento-sdk/src/main/java/momento/sdk/internal/GrpcChannelOptions.java
```diff
@@ -0,0 +1,38 @@
...
+  // The default value for max_send_message_length is 4mb.  We need to increase this to 5mb in order
+  // to
+  // support cases where users have requested a limit increase up to our maximum item size of 5mb.
+  public static final int DEFAULT_MAX_MESSAGE_SIZE = 5_243_000; // bytes
+
+  public static final boolean DEFAULT_KEEPALIVE_WITHOUT_STREAM = true;
+  public static final int DEFAULT_KEEPALIVE_TIME_MS = 5000; // milliseconds
+  public static final int DEFAULT_KEEPALIVE_TIMEOUT_MS = 1000; // milliseconds
+
+  public static NettyChannelBuilder GrpcOptionsFromGrpcConfig(
```
anitarua, line 17: I'm not actually sure if it's idiomatic/acceptable to pass around builder objects like this, open to revising this in a different way if needed

cprice404: I think it's probably okay, I'd just change this signature a bit. You're not returning any grpc options, so I'd probably just call it something like `applyGrpcConfigurationToChannelBuilder`. (Note that in Java, method names always start with a lower case letter, unlike .NET.)

Also since this function is being called purely for side effects I don't think there's any reason to return the builder object back out at the end.

momento-sdk/src/main/java/momento/sdk/ScsControlGrpcStubsManager.java
```diff
@@ -26,16 +29,24 @@ final class ScsControlGrpcStubsManager implements AutoCloseable {
...
         NettyChannelBuilder.forAddress(credentialProvider.getControlEndpoint(), 443);
-    channelBuilder.useTransportSecurity();
-    channelBuilder.disableRetry();
+
+    // Override grpc config to disable keepalive for control clients
+    final GrpcConfiguration controlConfig =
+        configuration.getTransportStrategy().getGrpcConfiguration().withKeepAliveDisabled();
+
+    // set additional channel options (message size, keepalive, auth, etc)
+    channelBuilder = GrpcChannelOptions.GrpcOptionsFromGrpcConfig(controlConfig, channelBuilder);
```
cprice404, line 48: if you change this method sig the way I suggested below, I think this var will be able to go back to being `final`.

momento-sdk/src/main/java/momento/sdk/config/transport/GrpcConfiguration.java
```diff
@@ -4,32 +4,59 @@
...
 import javax.annotation.Nonnull;
+import momento.sdk.internal.GrpcChannelOptions;
 
 /** Abstracts away the gRPC configuration tunables. */
 public class GrpcConfiguration {
 
   private final Duration deadline;
   private final int minNumGrpcChannels;
+  private final int maxMessageSize;
+  private final boolean keepAliveWithoutCalls;
```
cprice404, line 15: Since, in Java, there is a sane language mechanism that we can use to distinguish between whether these have been explicitly set or not, we should use it (rather than overloading the meaning of default values like `0` as we were forced to do in golang).

In Java, the right way to define these three arguments is `Optional<Boolean>`, etc.

Then for the Lambda config you can set them to `Optional.empty()`, and in the method where you're translating them to grpc, you can do `if keepaliveWithoutCalls.isSet()` instead of `> 0` etc.

cprice404 (commented):
looking good, some minor comments, and would also like to get @pratik151192 or someone to review before we merge.

momento-sdk/src/main/java/momento/sdk/config/transport/GrpcConfiguration.java
```diff
@@ -3,33 +3,61 @@
...
+import java.util.OptionalInt;
 import javax.annotation.Nonnull;
+import momento.sdk.internal.GrpcChannelOptions;
 
 /** Abstracts away the gRPC configuration tunables. */
 public class GrpcConfiguration {
 
   private final Duration deadline;
   private final int minNumGrpcChannels;
+  private final OptionalInt maxMessageSize;
```
cprice404, line 15: interesting, i'd never seen `OptionalInt` before, only `Optional<Int>`. but this seems fine.

is there a reason not to make the boolean one Optional as well?

cprice404 (commented):
lgtm except one last nitpicky question

momento-sdk/src/main/java/momento/sdk/config/transport/GrpcConfiguration.java
anitarua: I was only able to get `OptionalInt` to compile (not `Optional<int>`) and there was no OptionalBoolean equivalent.

Did another quick search and found [this](https://schneide.blog/2019/04/23/javas-optionalint-et-al-versus-optionalt/) so I guess I just missed the fact that Java has both `int` and `Integer`. Will switch to using `Optional<Integer>` and `Optional<Boolean>` then.

cprice404: yeah, generics only work with the "boxed" data types, sorry I forgot to mention that.

cprice404 (approved):
lgtm. let's get either @pratik151192 or @nand4011 to review, just to make sure we're all aligned on the new API surface area, then 🚢 

nice

nand4011 (approved):

PR #354, opened 2024-03-14
chore: add cancelled errors to load-gen

pratik151192:
Have edited load-gens a couple times locally for cancelled might as well commit it

### DIFF
--- a/examples/cache/src/main/java/momento/client/example/LoadGenerator.java
+++ b/examples/cache/src/main/java/momento/client/example/LoadGenerator.java
@@ -42,6 +42,7 @@ public class LoadGenerator {
   private final LongAdder globalUnavailableCount = new LongAdder();
   private final LongAdder globalTimeoutCount = new LongAdder();
   private final LongAdder globalLimitExceededCount = new LongAdder();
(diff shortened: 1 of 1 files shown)


examples/cache/src/main/java/momento/client/example/LoadGenerator.java
```diff
@@ -244,7 +251,7 @@ public static void main(String[] args) throws InterruptedException {
     // may increase.
     // Note: You are likely to see degraded performance if you increase this above 50
     // and observe elevated client-side latencies.
-    final int numberOfConcurrentRequests = 50;
+    final int numberOfConcurrentRequests = 10;
```
nand4011, line 254: Did you intend to change this?

nand4011 (approved):

examples/cache/src/main/java/momento/client/example/LoadGenerator.java
pratik151192: yeah because max TPS was 50 :) so 1:1 here isn't needed and also is a bit misleading/confusing

PR #357, opened 2024-05-28
feat: add get batch and set batch

nand4011:
Add the get batch and set batch operations. They are blocking internally so they need blocking stubs.

Update the protos to the newest version.


momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -262,6 +264,18 @@ public CompletableFuture<GetResponse> get(String cacheName, String key) {
...
 
+  /**
+   * Get the cache values stored for the given keys.
+   *
+   * @param cacheName Name of the cache to get the item from.
+   * @param keys The keys to get.
+   * @return Future with {@link GetBatchResponse} containing the status of the get batch operation
+   *     and the associated value data.
+   */
+  public CompletableFuture<GetBatchResponse> getBatch(String cacheName, Iterable<String> keys) {
```
nand4011, line 275: The gRPC calls for get batch and set batch are blocking, so the user-facing calls are only pretend async. I did that to keep the API consistent, but I can change it if we want to ensure users know these calls will block.

momento-sdk/src/main/java/momento/sdk/ScsDataClient.java
```diff
@@ -1420,6 +1455,46 @@ public void onFailure(@Nonnull Throwable e) {
     return returnFuture;
   }
 
+  private GetBatchResponse sendGetBatch(String cacheName, List<ByteString> keys) {
+    checkCacheNameValid(cacheName);
+
+    final Metadata metadata = metadataWithCache(cacheName);
+    try {
+      final Iterator<_GetResponse> responses =
+          attachMetadataBlocking(scsDataGrpcStubsManager.getBlockingStub(), metadata)
```
cprice404, line 1464: So I don't really get how all these stubs work, but if you look at the topic client, it's calling `newStub` rather than `newBlockingStub` or `newFutureStub`:

https://github.com/momentohq/client-sdk-java/blob/7bba262d18275a4375da454ddbbe7f9645d36e72/momento-sdk/src/main/java/momento/sdk/ScsTopicGrpcStubsManager.java#L32

Then it is able to make a call to an async streaming API (subscribe):

https://github.com/momentohq/client-sdk-java/blob/51ca3b47190053e52fb7e7a47d1b9a0f86cf574b/momento-sdk/src/main/java/momento/sdk/ScsTopicClient.java#L160

It uses this observer thing instead of getting back a response directly.

I see in the generated ScsGrpc class that there are three classes: `ScsFutureStub`, `ScsBlockingStub`, and `ScsStub`. The `ScsStub` (which, unfortunately, we don't have any instances of at the moment) does have what appear to be async-based implementations of the getBatch and setBatch, that use this observer pattern.

It seems like that's what we should be using?

I don't think this is urgent, so I might be tempted to just add some comments here and save this for a future optimization if needed... or just put this PR on hold for now... depending on how much work you think this would be.

My one concern about merging this as-is would be that it might be easy for someone to tie up their main threads by trying to do many concurrent getbatch/setbatch calls, and they would assume that we would behave better than we actually will because of our signatures. wdyt?

nand4011: You're right, I missed the ScsStub implementations. I switched getBatch/setBatch to use that instead. It would be nice if we could expose the stream to the user, but that is a bigger ux change. We can save it for the Kotlin version :-)

cprice404: I think the idea was always to expose these as a fully populated `List` for v0; we can add a `getBatchStreaming` in the future that accepts a callback if people want to interact with the items as they arrive.

cprice404 (approved):

PR #358, opened 2024-06-04
feat: persistent store spike

malandis:
Adds a POC Persistent Store client, complete with client interface,
client, driver program, and response types.

The design focus is around the get request and set response. Since the
persistent store allows users to store and retrieve data of various
types, we must handle those cases carefully.

To store data, we overload the `set` method. There is a `set` overload
for each of the supported value types.

To retrieve data, we have a single `GetResponse` class. This class
stores the value as an object and the type as reported from the server
as an enumerated field. Users can inspect the type of the value by
calling the accessor `getType` and then calling the corresponding
value accessor for strongly typed access (`getValueAsString`,
`getValueAsByteArray`, ...).

In the event a user calls the wrong method for the underlying type, we
throw an exception.

We add a driver program to demonstrate usage in
`PersistentStoreDriver`. This shows how to exhaustively check a get
response when the underlying type is unknown. We also show what to do
when the type is expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,28 @@
+.PHONY: clean build test format lint precommit help
+
+## Clean the project
+clean:
+	./gradlew clean
+
+## Build the project
+build:
+	./gradlew build
+
+## Run the tests
+test:
+	./gradlew test
+
+## Format the code
+format:
+	./gradlew spotlessApply
+
+## Lint the code
+lint:
+	./gradlew spotlessCheck
+
+## Run the precommit checks
+precommit: format lint test
+
+# See <https://gist.github.com/klmr/575726c7e05d8780505a> for explanation.
+help:
+	@echo "$$(tput bold)Available rules:$$(tput sgr0)";echo;sed -ne"/^## /{h;s/.*//;:d" -e"H;n;s/^## //;td" -e"s/:.*//;G;s/\\n## /---/;s/\\n/ /g;p;}" ${MAKEFILE_LIST}|LC_ALL='C' sort -f|awk -F --- -v n=$$(tput cols) -v i=19 -v a="$$(tput setaf 6)" -v z="$$(tput sgr0)" '{printf"%s%*s%s ",a,-i,$$1,z;m=split($$2,w," ");l=n-i;for(j=1;j<=m;j++){l-=length(w[j])+1;if(l<= 0){l=n-i-length(w[j])-1;printf"\n%*s ",-i," ";}printf"%s ",w[j];}printf"\n";}'|more $(shell test $(shell uname) == Darwin && echo '-Xr')
--- a/momento-sdk/src/main/java/PersistentStoreDriver.java
+++ b/momento-sdk/src/main/java/PersistentStoreDriver.java
@@ -1,0 +1,90 @@
+import momento.sdk.IPersistentStoreClient;
+import momento.sdk.PersistentStoreClient;
+import momento.sdk.exceptions.ClientSdkException;
+import momento.sdk.responses.persistent_store.control.CreatePersistentStoreResponse;
+import momento.sdk.responses.persistent_store.control.ListPersistentStoresResponse;
+import momento.sdk.responses.persistent_store.data.GetResponse;
+
+public class PersistentStoreDriver {
+  public static void main(String[] args) {
+    // Instantiating the client will be largely the same as compared to cache/topics.
+    IPersistentStoreClient client = new PersistentStoreClient();
+
+    // Create a store
+    CreatePersistentStoreResponse createResponse = client.createStore("myStore").join();
+    if (createResponse instanceof CreatePersistentStoreResponse.Success) {
+      System.out.println("Store created successfully");
+    } else {
+      // Would inspect for Already exists, etc.
+      System.out.println("Error creating store");
+    }
+
+    // List all stores
+    final ListPersistentStoresResponse listResponse = client.listStores().join();
+    if (listResponse instanceof ListPersistentStoresResponse.Success) {
+      ListPersistentStoresResponse.Success success =
+          (ListPersistentStoresResponse.Success) listResponse;
+      System.out.println("Stores:");
+      success.getStores().forEach(store -> System.out.println("- Store: " + store.getName()));
+    } else if (listResponse instanceof ListPersistentStoresResponse.Error) {
+      System.out.println(
+          "Error listing stores: "
+              + ((ListPersistentStoresResponse.Error) listResponse).getMessage());
+    } else {
+      System.out.println("Unknown response type");
+    }
+
+    // Set a value in the store
+    client.set("myStore", "myKey", "myValue").join();
+
+    // Get the value from the store
+    final GetResponse stringResponse = client.get("myStore", "string").join();
+    printGenericGetResponse(stringResponse);
+
+    final GetResponse doubleResponse = client.get("myStore", "double").join();
+    printGenericGetResponse(doubleResponse);
+
+    if (stringResponse instanceof GetResponse.Success) {
+      GetResponse.Success success = (GetResponse.Success) doubleResponse;
+      System.out.println("Value type for \"double\" was: " + success.getType());
+      System.out.println("Value: " + success.getValueAsDouble());
+
+      // Exception
+      try {
+        System.out.println("Value: " + success.getValueAsString());
+      } catch (ClientSdkException e) {
+        System.out.println("Error trying to access the value of \"double\" as a string: " + e.getMessage());
+      }
+    } else if (stringResponse instanceof GetResponse.Error) {
+      GetResponse.Error error = (GetResponse.Error) doubleResponse;
+      System.out.println("Error: " + error.getMessage());
+    } else {
+      System.out.println("Unknown response type");
+    }
+  }
+
+  /** Demos how to handle a get response with an unknown type. */
+  public static void printGenericGetResponse(GetResponse response) {
+    if (response instanceof GetResponse.Success) {
+      final GetResponse.Success success = (GetResponse.Success) response;
+      // IntelliJ has an option to enforce enum switch completeness
+      switch (success.getType()) {
+        case STRING:
+          System.out.println("Value: " + success.getValueAsString());
+          break;
+        case BYTE_ARRAY:
+          System.out.println("Value: " + new String(success.getValueAsByteArray()));
+          break;
+        case LONG:
+          System.out.println("Value: " + success.getValueAsLong());
+          break;
+        case DOUBLE:
+          System.out.println("Value: " + success.getValueAsDouble());
+          break;
+      }
+    } else if (response instanceof GetResponse.Error) {
+      final GetResponse.Error error = (GetResponse.Error) response;
+      System.out.println("Error: " + error.getMessage());
+    }
+  }
+}
--- a/momento-sdk/src/main/java/momento/sdk/IPersistentStoreClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/IPersistentStoreClient.java
@@ -1,0 +1,31 @@
+package momento.sdk;
+
+import java.util.concurrent.CompletableFuture;
+import momento.sdk.responses.persistent_store.control.CreatePersistentStoreResponse;
+import momento.sdk.responses.persistent_store.control.DeletePersistentStoreResponse;
+import momento.sdk.responses.persistent_store.control.ListPersistentStoresResponse;
+import momento.sdk.responses.persistent_store.data.DeleteResponse;
+import momento.sdk.responses.persistent_store.data.GetResponse;
+import momento.sdk.responses.persistent_store.data.SetResponse;
+
+public interface IPersistentStoreClient {
+  /** Control operations */
+  CompletableFuture<CreatePersistentStoreResponse> createStore(String storeName);
+
+  CompletableFuture<DeletePersistentStoreResponse> deleteStore(String storeName);
+
+  CompletableFuture<ListPersistentStoresResponse> listStores();
+
+  /** Data operations */
+  CompletableFuture<GetResponse> get(String storeName, String key);
+
+  CompletableFuture<SetResponse> set(String storeName, String key, byte[] value);
+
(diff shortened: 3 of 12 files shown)


momento-sdk/src/main/java/PreviewStorageClientDriver.java
```diff
@@ -0,0 +1,89 @@
...
+import momento.sdk.PreviewStorageClient;
+import momento.sdk.exceptions.ClientSdkException;
+import momento.sdk.responses.storage.control.CreateStoreResponse;
+import momento.sdk.responses.storage.control.ListStoresResponse;
+import momento.sdk.responses.storage.data.GetResponse;
+
+public class PreviewStorageClientDriver {
+  public static void main(String[] args) {
+    // Instantiating the client will be largely the same as compared to cache/topics.
+    IPreviewStorageClient client = new PreviewStorageClient();
```
cprice404, line 11: don't we need config/creds here?

momento-sdk/src/main/java/PreviewStorageClientDriver.java
```diff
@@ -0,0 +1,89 @@
...
+
+    // Get the value from the store
+    final GetResponse stringResponse = client.get("myStore", "string").join();
+    printGenericGetResponse(stringResponse);
+
+    final GetResponse doubleResponse = client.get("myStore", "double").join();
+    printGenericGetResponse(doubleResponse);
+
+    if (stringResponse instanceof GetResponse.Success) {
+      GetResponse.Success success = (GetResponse.Success) doubleResponse;
```
cprice404, line 46: you're doing `instanceof` on `stringResponse` and then casting the `doubleResponse`?

momento-sdk/src/main/java/PreviewStorageClientDriver.java
```diff
@@ -0,0 +1,89 @@
...
+    client.set("myStore", "myKey", "myValue").join();
+
+    // Get the value from the store
+    final GetResponse stringResponse = client.get("myStore", "string").join();
+    printGenericGetResponse(stringResponse);
+
+    final GetResponse doubleResponse = client.get("myStore", "double").join();
+    printGenericGetResponse(doubleResponse);
+
+    if (stringResponse instanceof GetResponse.Success) {
```
cprice404, line 45: This might be a @nand4011 question but I'm trying to remember if we found a nicer way to do these due to pattern matching improvements in later versions of Java, and whether we have a way to make that visible to users?

momento-sdk/src/main/java/PreviewStorageClientDriver.java
```diff
@@ -0,0 +1,89 @@
...
+
+      // Exception
+      try {
+        System.out.println("Value: " + success.getValueAsString());
+      } catch (ClientSdkException e) {
+        System.out.println(
+            "Error trying to access the value of \"double\" as a string: " + e.getMessage());
+      }
+    } else if (stringResponse instanceof GetResponse.Error) {
+      GetResponse.Error error = (GetResponse.Error) doubleResponse;
```
cprice404, line 58: same comment here about using a different var in the condition vs the cast

momento-sdk/src/main/java/PreviewStorageClientDriver.java
malandis: For now I've written requests, responses, the interface, and integration tests. I was going to save connecting and the credentials for when the service is live. Shall I do that now?

momento-sdk/src/main/java/momento/sdk/PreviewStorageClient.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  public CompletableFuture<ListStoresResponse> listStores() {
+
+    return CompletableFuture.completedFuture(new ListStoresResponse.Success());
+  }
+
+  /** Data operations */
+  public CompletableFuture<GetResponse> get(String storeName, String key) {
+    if (key.equalsIgnoreCase("byte array")) {
```
cprice404, line 29: You can go ahead and wirte this stuff up to the protos, right?

momento-sdk/src/main/java/momento/sdk/responses/storage/data/GetResponse.java
```diff
@@ -0,0 +1,86 @@
...
+      ensureCorrectTypeOrThrowException(ValueType.STRING, valueType);
+      return (String) value;
+    }
+
+    public long getValueAsLong() {
+      ensureCorrectTypeOrThrowException(ValueType.LONG, valueType);
+      return (long) value;
+    }
+
+    public double getValueAsDouble() {
```
malandis, line 54: For consistency with the cache client, I could also rename these `valueAsByteArray`, `valueAsString`, ... etc though I feel `get ... As ...` is easier to read/more idiomatic to Java.

cprice404 (commented):
this is looking good to me! left a few comments/questions. Would like to have @nand4011 review as well.

momento-sdk/src/main/java/PreviewStorageClientDriver.java
malandis: That was a typo

momento-sdk/src/main/java/PreviewStorageClientDriver.java
malandis: There is, but since this file is in the `momento.sdk` part of the repo, it's Java 8. The syntax is identical to the one in .NET: https://github.com/momentohq/client-sdk-java/blob/main/examples/cache/src/main/java/momento/client/example/BasicExample.java#L68-L76

momento-sdk/src/main/java/PreviewStorageClientDriver.java
malandis: Another typo. Good 👁️

momento-sdk/src/main/java/momento/sdk/PreviewStorageClient.java
malandis: I added these mocks so the driver program was usable.

momento-sdk/src/main/java/momento/sdk/responses/storage/data/GetResponse.java
malandis: Thoughts @cprice404 ?

momento-sdk/src/main/java/PreviewStorageClientDriver.java
malandis: Added in [14c4367](https://github.com/momentohq/client-sdk-java/pull/358/commits/14c4367eef290a467d15bd096c191323c2d932c1)

momento-sdk/src/main/java/momento/sdk/responses/storage/data/GetResponse.java
cprice404: i think i probably agree with your assessment about which would be more idiomatic, but i think consistency has to win out here. thanks for checking.

momento-sdk/src/main/java/PreviewStorageClientDriver.java
cprice404: okay so the idea is that this file can move into the examples dir once there is a release, and we will have the option of showing the more modern patterns?

momento-sdk/src/main/java/momento/sdk/PreviewStorageClient.java
cprice404: that makes sense, just checking that the plan is to fill them in as later commits on this pr.

momento-sdk/src/main/java/momento/sdk/config/StorageConfigurations.java
```diff
@@ -0,0 +1,123 @@
...
+      return Laptop.v1();
+    }
+
+    /**
+     * Provides v1 configuration for a laptop environment. This configuration is guaranteed not to
+     * change in future releases of the Momento Java SDK.
+     *
+     * @return the v1 laptop configuration
+     */
+    public static StorageConfiguration v1() {
```
cprice404, line 44: let's maybe avoid putting any v1s into these for now, so we have some wiggle room to decide what we want default retries/timeouts to look like?

cprice404: so just rename this to `latest` and get rid of the `latest` above?

momento-sdk/src/main/java/momento/sdk/config/StorageConfigurations.java
```diff
@@ -0,0 +1,123 @@
...
+      final TransportStrategy transportStrategy =
+          // TODO
+          new StaticTransportStrategy(new GrpcConfiguration(Duration.ofMillis(15000)));
+      final RetryStrategy retryStrategy =
+          new FixedCountRetryStrategy(DEFAULT_MAX_RETRIES, new DefaultRetryEligibilityStrategy());
```
cprice404, line 49: Is this eligibility strategy going to do anything here? It's checking for request types that only exist in the Cache APIs, right?

malandis: We can add our own that defines types for the storage client.

momento-sdk/src/main/java/momento/sdk/config/StorageConfigurations.java
malandis: Removed in [5fb1f90](https://github.com/momentohq/client-sdk-java/pull/358/commits/5fb1f90690c7d1764b36c95ede12a007bcf45c29)

momento-sdk/src/main/java/momento/sdk/PreviewStorageClient.java
malandis: I wrote the data and control clients but left this as is for test driving.

momento-sdk/src/main/java/momento/sdk/responses/storage/data/GetResponse.java
malandis: Renamed in [451f649](https://github.com/momentohq/client-sdk-java/pull/358/commits/451f6492c9f755fade41da4d26dc37eda9928352) to be idiomatic with cache response accessors.

malandis:
This is ready / pending end to end testing.

momento-sdk/src/main/java/PreviewStorageClientDriver.java
nand4011: The patterns we use in the examples are were introduced in Java 14.

momento-sdk/src/main/java/momento/sdk/StorageDataClient.java
```diff
@@ -0,0 +1,253 @@
...
+        // Execute on same thread that called execute on CompletionStage
+        MoreExecutors.directExecutor());
+
+    return returnFuture;
+  }
+
+  private CompletableFuture<SetResponse> sendSet(String key, _StoreValue value) {
+    // Submit request to non-blocking stub
+    final ListenableFuture<_StoreSetResponse> rspFuture =
+        attachMetadata(storageDataGrpcStubsManager.getStub(), new Metadata())
```
nand4011, line 172: Don't you need the store name here?

nand4011 (commented):
LGTM after set takes a store name.

malandis:
All set

cprice404 (commented):
just clearing this off of y review list for now, please add me again when it's ready for 👀 !

malandis:
In a follow up PR we will refactor the errors to provide callers with a first-class way to distinguish `store not found` and `key not found`. We will also clean up the messages to refer to storage vs cache.

malandis:
I will document the `StorageValue` class and `GetResponse` methods that were added once we agree the API looks good.

momento-sdk/src/main/java/PreviewStorageClientDriver.java
cprice404: @nand4011 btw I was meaning to ask you where is the spot where you configured the examples to use 14? I couldn't find it. I was considering experimenting with making some of them use 21.

momento-sdk/src/intTest/java/momento/sdk/storage/DataTests.java
```diff
@@ -0,0 +1,156 @@
...
+    final long value = 42L;
+
+    // Successful Set
+    final PutResponse putResponse = client.put(storeName, key, value).join();
+    assertThat(putResponse).isInstanceOf(PutResponse.Success.class);
+
+    // Successful Get
+    final GetResponse getResponse = client.get(storeName, key).join();
+    assertThat(getResponse).isInstanceOf(GetResponse.Success.class);
+    assertThat(getResponse.success().get().value().get().getLong().get()).isEqualTo(value);
```
cprice404, line 83: It's a little long winded but I think each stop in the chain has a purpose. I think we will get asked why we can't remove one link in the chain and either have `success()` directly return `Optional<StorageValue>`, or just expose `.value() -> Optional<StorageValue>` directly on the response parent class. I think in either of those cases we'd be forced to throw an exception if it was not a `Success`. So I think I'm probably leaning towards keeping this as-is for now and getting feedback from others during the bug bash. 

@nand4011 wdyt?

momento-sdk/src/main/java/momento/sdk/responses/storage/data/GetResponse.java
```diff
@@ -0,0 +1,77 @@
+package momento.sdk.responses.storage.data;
```
cprice404, line 1: it feels a tiny bit weird that these have `.data` in the package name when the cache ones do not

malandis: I can remove the extra level of nesting. It's too bad there is no way to roll up the names as in other languages.

malandis: Done in [fbcda96](https://github.com/momentohq/client-sdk-java/pull/358/commits/fbcda961d89ea39425b8d35cea5791018d2ced17)

cprice404:
I would like to get @nand4011 opinion on this when he has time but I'm going to approve to unblock Michael from following up with smaller PRs.

cprice404 (approved):

PR #359, opened 2024-06-14
feat: onetime headers for agent and runtime-version

rishtigupta:
## Issue:
- https://github.com/momentohq/dev-eco-issue-tracker/issues/891
- https://github.com/momentohq/client-sdk-java/issues/372

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/ScsClientBase.java
+++ b/momento-sdk/src/main/java/momento/sdk/ScsClientBase.java
@@ -9,9 +9,29 @@ import io.grpc.stub.MetadataUtils;
 abstract class ScsClientBase extends ClientBase {
   private static final Metadata.Key<String> CACHE_NAME_KEY =
       Metadata.Key.of("cache", ASCII_STRING_MARSHALLER);
+  private static final Metadata.Key<String> SDK_AGENT_KEY =
+          Metadata.Key.of("Agent", ASCII_STRING_MARSHALLER);
+  private static final Metadata.Key<String> RUNTIME_VERSION_KEY =
+          Metadata.Key.of("Runtime-Version", ASCII_STRING_MARSHALLER);
+  private boolean hasSentOneTimeHeaders = false;
 
   protected Metadata metadataWithCache(String cacheName) {
-    return metadataWithItem(CACHE_NAME_KEY, cacheName);
+    if (hasSentOneTimeHeaders) {
+      Metadata metadata = new Metadata();
+      metadata.put(CACHE_NAME_KEY, cacheName);
+      return metadata;
+    }
+    hasSentOneTimeHeaders = true;
+    String sdkVersion =
+            String.format("java:cache:%s", this.getClass().getPackage().getImplementationVersion());
+    String runtimeVer =
+            System.getProperty("java.vendor") + ", " + System.getProperty("java.version");
+
+    Metadata metadata = new Metadata();
+    metadata.put(CACHE_NAME_KEY, cacheName);
+    metadata.put(SDK_AGENT_KEY, sdkVersion);
+    metadata.put(RUNTIME_VERSION_KEY, runtimeVer);
+    return metadata;
   }
 
   protected <S extends AbstractStub<S>> S attachObservableMetadata(S stub, Metadata metadata) {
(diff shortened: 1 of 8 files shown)


bruuuuuuuce:
before we merge this, lets finish the discussion whether or not we want to add the client name in the agent header

momento-sdk/src/main/java/momento/sdk/ScsClient.java
```diff
@@ -19,11 +19,27 @@ abstract class ScsClient implements AutoCloseable {
...
-    metadata.put(CACHE_NAME_KEY, cacheName);
+    if (hasSentOneTimeHeaders) {
+      Metadata metadata = new Metadata();
+      metadata.put(CACHE_NAME_KEY, cacheName);
+      return metadata;
+    }
+    hasSentOneTimeHeaders = true;
+    String sdkVersion =
+        String.format("java:%s", this.getClass().getPackage().getImplementationVersion());
+    String runtimeVer = System.getProperty("java.version");
```
nand4011, line 37: Is it worth including the vendor `System.getProperty("java.vendor")` as well, so we know which jdk is being used? The non-android runtime version string I'm using for kotlin at the moment is `val runtimeVersion = "$javaVendor-$javaVersion, Kotlin $kotlinVersion"`
cc @cprice404

momento-sdk/src/main/java/momento/sdk/UserHeaderInterceptor.java
```diff
@@ -35,6 +40,7 @@ public void start(Listener<RespT> listener, Metadata metadata) {
         metadata.put(AUTH_HEADER_KEY, tokenValue);
         if (!isUserAgentSent) {
           metadata.put(SDK_AGENT_KEY, sdkVersion);
+          metadata.put(RUNTIME_VERSION_KEY, runtimeVer);
```
bruuuuuuuce, line 43: just wondering, it looks like we are setting these in 2 different places? What is the reasoning behind that?

momento-sdk/src/main/java/momento/sdk/ScsClientBase.java
```diff
@@ -9,9 +9,29 @@
...
   protected Metadata metadataWithCache(String cacheName) {
-    return metadataWithItem(CACHE_NAME_KEY, cacheName);
+    if (hasSentOneTimeHeaders) {
+      Metadata metadata = new Metadata();
+      metadata.put(CACHE_NAME_KEY, cacheName);
+      return metadata;
+    }
+    hasSentOneTimeHeaders = true;
+    String sdkVersion =
+        String.format("java:cache:%s", this.getClass().getPackage().getImplementationVersion());
```
nand4011, line 26: Why is this needed if the user header interceptor code exists?

momento-sdk/src/main/java/momento/sdk/UserHeaderInterceptor.java
rishtigupta: I noticed that too. But this is the pattern we were using: Clients use the "metadataWithCache" from ScsClientBase class to create headers, whereas the stub managers use the UseHeaderInterceptor class to create the headers.

nand4011: I think you only need to add the headers once. It might make more sense to use the metadataWithCache style that you added and get rid of this interceptor, because `isUserAgentSent` in this class is static, so it will only add the headers to the first momento client that executes a call. If you see `java:cache:1.0.0`, you'll never see `java:store:1.0.0`.

momento-sdk/src/main/java/momento/sdk/UserHeaderInterceptor.java
```diff
@@ -13,16 +13,21 @@
...
-      Metadata.Key.of("Agent", ASCII_STRING_MARSHALLER);
+      Metadata.Key.of("agent", ASCII_STRING_MARSHALLER);
+  private static final Metadata.Key<String> RUNTIME_VERSION_KEY =
+      Metadata.Key.of("runtime-version", ASCII_STRING_MARSHALLER);
   private final String tokenValue;
-  private final String sdkVersion =
-      String.format("java:%s", this.getClass().getPackage().getImplementationVersion());
-  private static volatile boolean isUserAgentSent = false;
+  private final String sdkVersion;
+  String runtimeVer = System.getProperty("java.vendor") + ", " + System.getProperty("java.version");
```
nand4011, line 23: Can you make this private final to match the sdkVersion? And define them both in the same place.

PR #360, opened 2024-06-20
refactor: introduce `Store(Item)?NotFoundException` classes

malandis:
This PR detects when a `NOT_FOUND` gRPC error was specific to the
storage client or not. For the storage client `NOT_FOUND`, there are
two cases: the store was not found and the item was not found.

To distinguish these two cases we use string matching on the
grpcException message.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/exceptions/CacheServiceExceptionMapper.java
+++ b/momento-sdk/src/main/java/momento/sdk/exceptions/CacheServiceExceptionMapper.java
@@ -44,12 +44,17 @@ public final class CacheServiceExceptionMapper {
     if (e instanceof io.grpc.StatusRuntimeException) {
       final StatusRuntimeException grpcException = (StatusRuntimeException) e;
       final Status.Code statusCode = grpcException.getStatus().getCode();
+      final Metadata trailers = grpcException.getTrailers();
 
       final MomentoTransportErrorDetails errorDetails =
           new MomentoTransportErrorDetails(
               new MomentoGrpcErrorDetails(statusCode, grpcException.getMessage(), metadata));
 
-      final String errorCause = grpcException.getMessage();
+      String errorCause = trailers.get(Metadata.Key.of("err", Metadata.ASCII_STRING_MARSHALLER));
+      if (errorCause == null) {
+        // TODO remove once control service is updated to send "err" in metadata
+        errorCause = grpcException.getMessage();
+      }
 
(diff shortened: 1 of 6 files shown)


malandis:
Not terribly important or urgent: we may consider organizing the various `not found` exception classes as follows:
- a base class `ResourceNotFoundException`, with subclasses
- `CacheResourceNotFoundException` and `StoreResourceNotFoundException`
-  the latter derives `StoreItemNotFoundException` and `StoreNotFoundException`

nand4011 (approved):
LGTM. We'll have to be careful about building an exception hierarchy that includes cache exceptions, since we could technically introduce breaking changes if a call starts returning a different error type than before.

momento-sdk/src/main/java/momento/sdk/exceptions/CacheServiceExceptionMapper.java
```diff
@@ -44,11 +44,18 @@ public static SdkException convert(Throwable e, Metadata metadata) {
...
 
       final MomentoTransportErrorDetails errorDetails =
           new MomentoTransportErrorDetails(
               new MomentoGrpcErrorDetails(statusCode, grpcException.getMessage(), metadata));
 
+      String errorCause = trailers.get(Metadata.Key.of("err", Metadata.ASCII_STRING_MARSHALLER));
+      if (errorCause == null) {
+        // TODO remove once control service is updated to send "err" in metadata
+        errorCause = grpcException.getMessage();
+      }
```
malandis, line 57: Will update this once the server is up to date.

momento-sdk/src/main/java/momento/sdk/exceptions/CacheServiceExceptionMapper.java
```diff
@@ -75,8 +82,14 @@ public static SdkException convert(Throwable e, Metadata metadata) {
...
-          return new NotFoundException(grpcException, errorDetails);
-
+          if (errorCause.contains("element_not_found")) {
+            return new StoreItemNotFoundException(grpcException, errorDetails);
+            // TODO change once control service is updated to send "Store with name" in metadata
+          } else if (errorCause.contains("Store with name")) {
+            return new StoreNotFoundException(grpcException, errorDetails);
+          } else {
+            return new NotFoundException(grpcException, errorDetails);
+          }
```
malandis, line 92: Will update these once the server is up to date.

nand4011 (approved):

PR #361, opened 2024-06-20
refactor: rename `StorageValueType` to `StorageItemType`

malandis:
To align with the cache naming, we rename this to `StorageItemType`. A
future RPC could be `getItemType` and the name will align.


momento-sdk/src/main/java/momento/sdk/responses/storage/StorageValue.java
```diff
@@ -4,56 +4,56 @@
 
 public class StorageValue {
```
nand4011, line 5: Should this be `StorageItem`?

nand4011 (approved):

momento-sdk/src/main/java/momento/sdk/responses/storage/StorageValue.java
malandis: This only has the value part of the item (item = (key, value)).

PR #363, opened 2024-06-21
docs: document storage response types more thoroughly

malandis:
Adds more documentation for storage response types, especially the get
response and storage value.


rishtigupta (approved):
🚀

PR #366, opened 2024-06-21
chore: add to string implementations to relevant storage responses

malandis:
`GetResponse` and `ListStoresResponse` should have meaningful to
string implementations. This PR adds them.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/responses/storage/ListStoresResponse.java
+++ b/momento-sdk/src/main/java/momento/sdk/responses/storage/ListStoresResponse.java
@@ -17,6 +17,11 @@ public interface ListStoresResponse {
     public List<StoreInfo> getStores() {
       return stores;
     }
+
+    @Override
+    public String toString() {
+      return super.toString() + ": stores: " + stores;
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


momento-sdk/src/main/java/momento/sdk/responses/storage/ListStoresResponse.java
```diff
@@ -17,6 +17,11 @@ public Success(List<StoreInfo> stores) {
     public List<StoreInfo> getStores() {
       return stores;
     }
+
+    @Override
+    public String toString() {
+      return super.toString() + ": stores: " + stores;
```
nand4011, line 23: Does this print the stores, or does it print the memory address of the list? I think the CacheListResponse has a toString that joins the cache names into a string before printing them.

malandis: I just added `StoreInfo::toString` and the result was this:

```text
momento.sdk.responses.storage.ListStoresResponse$Success@7e1f584d: stores: [momento.sdk.responses.storage.StoreInfo@7dff6d05: name: name-d3fa8790-dee2-4dc9-8d22-f91a0b4a293f, momento.sdk.responses.storage.StoreInfo@45d64d27: name: my-test-cache, momento.sdk.responses.storage.StoreInfo@34fe326d: name: name-9eb5a61b-7387-42a2-986b-86d1f926f644]
```

Should I modify that further?

nand4011: Works for me.

nand4011 (approved):

PR #369, opened 2024-06-24
fix: use server metadata for item not found

malandis:
Consumes the server metadata change from `element_not_found` to `item_not_found`.


bruuuuuuuce (approved):
:shipit:

PR #373, opened 2024-06-27
chore: add storage client builder

malandis:
Adds a storage client builder for convenience. The default status of
credential provider and the configuration are subject to change.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/PreviewStorageClientBuilder.java
+++ b/momento-sdk/src/main/java/momento/sdk/PreviewStorageClientBuilder.java
@@ -1,0 +1,55 @@
+package momento.sdk;
+
+import javax.annotation.Nonnull;
+import momento.sdk.auth.CredentialProvider;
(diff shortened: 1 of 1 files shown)


pratik151192 (approved):

momento-sdk/src/main/java/momento/sdk/PreviewStorageClientBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+    this.configuration = configuration;
+    return this;
+  }
+
+  /**
+   * Builds a PreviewStorageClient.
+   *
+   * @return the client.
+   */
+  public IPreviewStorageClient build() {
```
nand4011, line 49: Is this where we should warn about using a laptop config?

malandis: I made a follow up ticket to capture this and the related design work.

PR #375, opened 2024-06-28
feat: storage get response 2.0

malandis:
1. Split `GetResponse.Success` into `GetResponse.Found` and, in the event the item isn't found, `GetResponse.NotFound`. This sheds the optional on `value`.
2. Add a `valueWhenFound` shortcut accessor on the interface that returns a `MomentoOptional<StorageValue>`.
3. ~~Throw exceptions when `StorageValue::get*` are called on the wrong type. This sheds the `Optional` on those responses.~~ Return a `MomentoOptional` which allows for both optional behavior and custom error messages. A careful user can still switch on `StorageValue::getType` for exhaustiveness checks.
4. ~~Add `GetResponse::asFound` which returns `GetResponseFound` in the event of a `GetResponse.Found` otherwise throws an exception.~~ Will add this on demand per user feedback in order to keep the API slim.
5. Introduce `MomentoOptional` for the convenience of a parameter-less `orElseThrow` method. The `orElseThrow` method will throw a `ClientSdkException` when the optional is empty with an informative error message. This makes using `GetResponse::found` slightly less long winded.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/responses/storage/GetResponseFound.java
+++ b/momento-sdk/src/main/java/momento/sdk/responses/storage/GetResponseFound.java
@@ -1,5 +1,0 @@
-package momento.sdk.responses.storage;
-
-public interface GetResponseFound {
-  StorageValue value();
-}
--- a/momento-sdk/src/main/java/momento/sdk/responses/storage/GetResponseFound.java
+++ b/momento-sdk/src/main/java/momento/sdk/responses/storage/GetResponseFound.java
@@ -1,0 +1,5 @@
+package momento.sdk.responses.storage;
+
+public interface GetResponseFound {
+  StorageValue value();
+}
--- a/momento-sdk/src/test/java/momento/sdk/responses/storage/GetResponseTest.java
+++ b/momento-sdk/src/test/java/momento/sdk/responses/storage/GetResponseTest.java
@@ -1,6 +1,9 @@
 package momento.sdk.responses.storage;
 
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
 import io.grpc.Status;
+import momento.sdk.exceptions.ClientSdkException;
 import momento.sdk.exceptions.StoreNotFoundException;
 import momento.sdk.internal.MomentoGrpcErrorDetails;
 import momento.sdk.internal.MomentoTransportErrorDetails;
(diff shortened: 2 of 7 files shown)


malandis:
Leaving this as draft for not but open for design/ergonomics review.

momento-sdk/src/main/java/momento/sdk/responses/storage/GetResponseFound.java
```diff
@@ -0,0 +1,5 @@
+package momento.sdk.responses.storage;
+
+public interface GetResponseFound {
+  StorageValue value();
+}
```
kvcache, line 5: Is there value in this extra type or should the "found" case simply yield a StorageValue, which is the only thing a GetResponse should yield in the found case?

momento-sdk/src/test/java/momento/sdk/responses/storage/GetResponseTest.java
```diff
@@ -1,64 +1,67 @@
...
-    assert response.value().get().getByteArray().get().length == 4;
-    assert !response.value().get().getString().isPresent();
-    assert !response.value().get().getLong().isPresent();
-    assert !response.value().get().getDouble().isPresent();
+  public void testGetResponseFoundWorksOnTheRightType() {
+    GetResponse.Found response = GetResponse.Found.of(new byte[] {0, 1, 2, 3});
+    assert response.value().getByteArray().get().length == 4;
+    assertThrows(ClientSdkException.class, response.value().getString()::orElseThrow);
+    assertThrows(ClientSdkException.class, response.value().getLong()::orElseThrow);
+    assertThrows(ClientSdkException.class, response.value().getDouble()::orElseThrow);
```
kvcache, line 19: I see that in this found situation, even for the use case of orElseThrow, the presence of GetResponseFound means that value() line noise is required to get at the value.
I think when a user has gotten to a point in the type system where a response is known to have been found, they should be able to call getString().orElseThrow() directly.
Further, have you considered making life easier for the throwy person and using `get*` and `as*` methods instead? `getString()` could throw, while `asString()` could return your optionaly type.
This would make these lines instead look like
```suggestion
    // When found, a value is a StoreValue.
    StoreValue response = StoreValue.of(new byte[] {0, 1, 2, 3});
    assert response.asByteArray().get().length == 4; // could use getByteArray here instead
    assertThrows(ClientSdkException.class, response::getString);
    assertThrows(ClientSdkException.class, response::getLong);
    assertThrows(ClientSdkException.class, response::getDouble);
```

I love the exhaustive pattern as you have it, minus perhaps the GetResponseFound indirection. I think the throwy path could stand to be trimmed still further to give a user exactly what they expect or else an exception straight away. Happy to chat more about it if you like!

momento-sdk/src/main/java/momento/sdk/responses/storage/GetResponseFound.java
malandis: I had considered this as well. For the first cut I have kept the response class in parallel with the wire response (ie a `GetResponse` has a `value` which is `oneof` ...). This made more sense when we had `Success` vs `Error` and the `Success` was a found vs not found. That is no longer the case.

Right now there is no other data in the response message. There is no danger in rolling up the value accessors here as long as we avoid name conflicts in the future.

momento-sdk/src/test/java/momento/sdk/responses/storage/GetResponseTest.java
malandis: > I think when a user has gotten to a point in the type system where a response is known to have been found, they should be able to call getString().orElseThrow() directly.

Agreed and will consider this.

> Further, have you considered making life easier for the throwy person and using get* and as* methods instead? getString() could throw, while asString() could return your optionaly type.

Yes I had this in a previous commit. We agreed internally to keep it slim for now and to add a parallel track of throwy methods per customer feedback.

cprice404: +1 to keeping slim for now. We can 'throwy' versions of things in the future if there is interest.

cprice404 (commented):
@malandis ping me when you're ready for another review based on our earlier discussion

cprice404 (approved):

cprice404:
@kvcache we spent some time discussing / playing around with ideas based on your feedback (thank you) and decided to stick with the possibility of adding the "throw-y" versions in the future. We did change the happy-path code route to expose `valueWhenFound` to give access to the value directly, rather than having to go through the `found()` chain. This sheds one level of accessor calls and puts us pretty much on par with SDKs like dynamo (though with more null safety).

We are going to go ahead and merge and release these changes so we'll at least be closer to the finish line if some users start testing this week.

If you have a better idea for that name, or other things you still feel strongly about, let us know. We would have preferred to call that just `value()` but the return type collides with the one from `GetResponse.Found.value()` so this was what we came up with.

PR #379, opened 2024-07-09
refactor: java cache and topics integration tests should use random cache names

malandis:
Refactors the cache and topics integration tests to use a random cache
name instead of reading from the environment. This follows the
JavaScript SDK pattern.

We have re-written the tests to use a base test class that sets up the
client, assigns a random cache name, creates the cache, and tears it
down at the end of tests.

Because a test class now re-uses the same cache, we have refactored
the tests to use random keys in each test case. These two changes have
fixed race conditions that have caused consistent test failures when
running locally.

Lastly the auth tests were not closing clients in the tests. Because
of this, warning messages appeared in the logs consistently. We have
fixed these problems by closing clients between reuse.

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/BaseTestClass.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/BaseTestClass.java
@@ -1,12 +1,15 @@
 package momento.sdk;
 
 import java.time.Duration;
+import java.util.UUID;
+
 import momento.sdk.auth.CredentialProvider;
 import momento.sdk.config.Configurations;
 import org.junit.jupiter.api.BeforeAll;
 
 public class BaseTestClass {
   public static final Duration FIVE_SECONDS = Duration.ofSeconds(5);
+  public static final Duration TEN_SECONDS = Duration.ofSeconds(10);
 
   public static final CredentialProvider credentialProvider =
       CredentialProvider.fromEnvVar("TEST_AUTH_TOKEN");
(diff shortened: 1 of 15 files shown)


momento-sdk/src/intTest/java/momento/sdk/MomentoBatchUtilsIntegrationTest.java
```diff
@@ -160,14 +147,17 @@ void testBatchGetWithStringKeys_RequestsMoreThanMaxConn() {
             .build();
 
     // test data with more keys than max concurrent requests
+    Vector<String> keys = new Vector<>();
```
nand4011, line 150:
```suggestion
    final List<String> keys = new ArrayList<>();
```
Is there a specific reason to use a Vector? I'm not sure I've ever seen one of those in use.

malandis: That's what I used in my intro to Java class 22 years ago. I can use `ArrayList` lol

momento-sdk/src/intTest/java/momento/sdk/BaseTestClass.java
```diff
@@ -1,35 +1,59 @@
...
   @BeforeAll
   static void beforeAll() {
-    if (System.getenv("TEST_AUTH_TOKEN") == null) {
-      throw new IllegalArgumentException(
-          "Integration tests require TEST_AUTH_TOKEN env var; see README for more details.");
-    }
-    if (System.getenv("TEST_CACHE_NAME") == null) {
-      throw new IllegalArgumentException(
-          "Integration tests require TEST_CACHE_NAME env var; see README for more details.");
+    credentialProvider = CredentialProvider.fromEnvVar("TEST_AUTH_TOKEN");
```
nand4011, line 22: Is it worth changing TEST_AUTH_TOKEN to MOMENTO_API_KEY in this pr? I thought we were slowly changing to that.

nand4011 (commented):
Looks good! Just a couple nits and I'll +1. Thanks for going through all those test classes.

momento-sdk/src/intTest/java/momento/sdk/BaseTestClass.java
malandis: I'm going to change the CI / env vars in a separate PR. This one is pretty busy already.

momento-sdk/src/intTest/java/momento/sdk/MomentoBatchUtilsIntegrationTest.java
malandis: Updated in [4eb3b71](https://github.com/momentohq/client-sdk-java/pull/379/commits/4eb3b719ac6a1e159d07a1e119369382a155b3f1)

nand4011 (approved):

PR #380, opened 2024-07-09
refactor: storage tests should use a single test store and base

malandis:
Refactors the storage tests to use a base class similar to the one
just introduced for cache. Handles store set up and tear down uniformly.

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/BaseStorageTestClass.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/BaseStorageTestClass.java
@@ -1,0 +1,55 @@
+package momento.sdk;
+
+import java.time.Duration;
+import java.util.UUID;
+import momento.sdk.auth.CredentialProvider;
+import momento.sdk.config.StorageConfigurations;
+import momento.sdk.responses.storage.CreateStoreResponse;
+import momento.sdk.responses.storage.DeleteStoreResponse;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+
+public class BaseStorageTestClass {
+  protected static final Duration TEN_SECONDS = Duration.ofSeconds(10);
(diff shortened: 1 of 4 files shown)


momento-sdk/src/main/java/momento/sdk/PreviewStorageClientBuilder.java
```diff
@@ -49,7 +49,7 @@ public PreviewStorageClientBuilder withConfiguration(
    *
    * @return the client.
    */
-  public IPreviewStorageClient build() {
+  public PreviewStorageClient build() {
```
malandis, line 52: I changed this since `PreviewStorageClient` is `Closeable`. In order to both use the builder and explicitly call `close` on the client, we should return the concrete class.

momento-sdk/src/intTest/java/momento/sdk/BaseStorageTestClass.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class BaseStorageTestClass {
+  protected static final Duration TEN_SECONDS = Duration.ofSeconds(10);
+  protected static CredentialProvider credentialProvider;
+  protected static PreviewStorageClient storageClient;
+  protected static String storeName;
+
+  @BeforeAll
+  static void beforeAll() {
+    credentialProvider = CredentialProvider.fromEnvVar("TEST_AUTH_TOKEN");
```
nand4011, line 20: This will need an update since your other pr got in first.

nand4011 (commented):
LGTM after the environment variable is updated.

momento-sdk/src/intTest/java/momento/sdk/BaseStorageTestClass.java
malandis: I will rebase that on the cache rename once that is through. This was the first PR.

nand4011 (approved):

PR #383, opened 2024-07-10
fix: storage SDK orElseThrow should take a supplier

cprice404:
The `orElseThrow` signature of Java Optionals requires a single
arg, `supplier`, that is a lambda function returning the exception
that the user wishes to have thrown if the optional is empty.
In our current storage client code we have an empty-args signature,
which would be surprising to Java users familiar with Optionals.
This commit updates the signature to accept a supplier.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/utils/MomentoOptional.java
+++ b/momento-sdk/src/main/java/momento/sdk/utils/MomentoOptional.java
@@ -1,6 +1,7 @@
 package momento.sdk.utils;
 
 import java.util.Optional;
+import java.util.function.Supplier;
 import momento.sdk.exceptions.ClientSdkException;
 
 /**
@@ -64,7 +65,7 @@ public class MomentoOptional<T> {
    * @return the value.
    */
   public T get() {
-    return this.orElseThrow();
+    return optional.orElseThrow(() -> new ClientSdkException(onEmptyExceptionMessage));
   }
 
   /**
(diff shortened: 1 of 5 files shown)


momento-sdk/src/main/java/momento/sdk/utils/MomentoOptional.java
```diff
@@ -64,16 +65,20 @@ public static <T> MomentoOptional<T> empty(String onEmptyExceptionMessage) {
    * @return the value.
    */
   public T get() {
-    return this.orElseThrow();
+    return optional.orElseThrow(() -> new ClientSdkException(onEmptyExceptionMessage));
```
malandis, line 68: 👍

momento-sdk/src/main/java/momento/sdk/utils/MomentoOptional.java
```diff
@@ -64,16 +65,20 @@ public static <T> MomentoOptional<T> empty(String onEmptyExceptionMessage) {
...
   /**
    * Gets the value of the optional or throws an exception if the optional is empty.
    *
    * @return the value.
    */
-  public T orElseThrow() {
-    return optional.orElseThrow(() -> new ClientSdkException(onEmptyExceptionMessage));
+  public <X extends Throwable> T orElseThrow(Supplier<? extends X> exceptionSupplier) throws X {
+    if (this.isEmpty()) {
+      throw exceptionSupplier.get();
```
malandis, line 78: 👍

malandis (approved):

PR #384, opened 2024-07-10
chore: add docs examples for PreviewStorageClient

cprice404:


examples/cache/src/main/java/momento/client/example/doc_examples/DocExamplesJavaAPIs.java
```diff
@@ -95,7 +95,7 @@ public static void example_API_CreateCache(CacheClient cacheClient) {
     if (response instanceof CacheCreateResponse.Success) {
       System.out.println("Cache 'test-cache' created");
     } else if (response instanceof CacheCreateResponse.Error error) {
-      if (error.getCause() instanceof AlreadyExistsException) {
+      if (error.getErrorCode() == MomentoErrorCode.ALREADY_EXISTS_ERROR) {
```
malandis, line 98: This SDK doesn't have the this as an alternative for the create cache response unfortunately.

malandis (approved):

examples/cache/src/main/java/momento/client/example/doc_examples/DocExamplesJavaAPIs.java
cprice404: yeah, i noticed that. I kind of want to fix that right now, do you have any opinion on it?

PR #387, opened 2024-08-30
docs: fix list fetch end index description

malandis:
Corrects the end index documentation on `listFetch` to indicate it is
exclusive, not inclusive.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/CacheClient.java
+++ b/momento-sdk/src/main/java/momento/sdk/CacheClient.java
@@ -1668,7 +1668,7 @@ public final class CacheClient implements AutoCloseable {
    * @param cacheName The cache containing the list.
    * @param listName The list to fetch.
    * @param startIndex Start index (inclusive) for fetch operation. Defaults to 0 if not provided.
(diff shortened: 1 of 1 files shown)


nand4011 (approved):

momento-sdk/src/main/java/momento/sdk/CacheClient.java
```diff
@@ -1668,7 +1668,7 @@ public CompletableFuture<ListConcatenateFrontResponse> listConcatenateFrontByteA
    * @param cacheName The cache containing the list.
    * @param listName The list to fetch.
    * @param startIndex Start index (inclusive) for fetch operation. Defaults to 0 if not provided.
-   * @param endIndex End index (inclusive) for fetch operation. Defaults to the end of the list if
+   * @param endIndex End index (exclusive) for fetch operation. Defaults to the end of the list if
```
nand4011, line 1671: I suppose technically it should be
```suggestion
   * @param endIndex End index (exclusive) for the fetch operation. Defaults to the end of the list if
```

PR #389, opened 2024-09-03
chore: Generify the scalar cache functions

nand4011:
Create a batch version of the existing generic execute grpc function and move setBatch and getBatch to it.

Switch the rest of the scalar calls to use the existing generic grpc execution function. This provides a single point where we can do things like limit concurrent calls.


cprice404 (approved):
consolidate

PR #390, opened 2024-09-03
fix: fix failing storage tests

nand4011:
Stop looking at error metadata to determine whether an already exists error is for a store or cache. We can check it when that is added on the server side.

Update a test that was checking that the server could support empty keys in the store client. That changed on the server side.


nand4011:
The build examples step can't go green because the released version uses the code no longer works.

malandis (approved):

PR #392, opened 2024-09-04
feat: add byte array for list methods on the list name

malandis:
Adds various overloads for cache `listFetch`,
`listConcatenateFrontByteArray`, and `listRetain` for the `listName`
as a byte array.


rishtigupta:
lgtm!
Do we want a case where we have `byte[] listname` and `string[] values` for  apis like `listConcatenate` ?

rishtigupta (approved):
:shipit:

PR #395, opened 2024-10-02
feat: Concurrency limiter

nand4011:
Adds concurrency limiting to the base client class in the style of https://github.com/momentohq/client-sdk-java/pull/385.

The concurrency limiting function is almost the same as the prototype and wraps the existing generic `executeGrpcFunction` and `executeGrpcBatchFunction` methods.

The PR is broken up into 3 commits:
The first adds the concurrency limiting fixed thread pool to the base client class, along with the constructor to configure it and the `close()` method to shut it down. It replaces the `close()` in the child classes with a new abstract `doClose()` so that there is still a close that the children are forced to implement.

The second adds the max concurrent requests integer to the transport strategy layer of the config and plumbs it through the cache data and control clients to the base client. The storage client config doesn't have it yet. It could have request limiting if we add the configuration option.

The third commit adds the request limiting function copied from the POC draft pr.

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/CacheClientTest.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/CacheClientTest.java
@@ -11,8 +11,10 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 import momento.sdk.auth.CredentialProvider;
 import momento.sdk.auth.StringCredentialProvider;
+import momento.sdk.config.Configuration;
 import momento.sdk.config.Configurations;
 import momento.sdk.exceptions.AuthenticationException;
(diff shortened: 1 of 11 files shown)


cprice404:
Thanks @nand4011  - we are still waiting on reproducer/perf testing data to validate this before I should dive in too deep to review, yes?

nand4011:
After fixing this bug https://github.com/momentohq/client-sdk-java/pull/396, allowing me to set timeouts lower than one second, the limiter provides benefits on startup when running a large number of concurrent calls.
When running at 1000 concurrent calls and 1000tps with a timeout of 100ms:
With no limiter, there are ~2000 timeouts on startup.
With a limit of 50 concurrent calls, there are ~60 timeouts.
With a limit of 10 concurrent calls, there are ~5 timeouts.
Throughput and latency are not affected.

cprice404 (approved):
bottleneck

PR #396, opened 2024-10-07
fix: set stub deadlines using millis instead of seconds

nand4011:
Set stub deadlines using milliseconds instead of seconds. Duration.getSeconds() truncates, so a client configuration with a timeout of less than one second, like the preconfigured LowLatency config, always has a deadline of 0, so no calls will ever succeed.

Remove the documentation from the topic stubs manager that states that it sets a deadline on the stub it returns.


rishtigupta (dismissed):
:shipit:

malandis (approved):

PR #399, opened 2024-11-01
chore: improve resource exhausted error message

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1052

Adds a factory method for `LimitExceededException` that takes in an error cause (`err` metadata or grpc error details string) and constructs a `LimitExceededException` with a message wrapper that includes the limit that was exceeded.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/exceptions/LimitExceededException.java
+++ b/momento-sdk/src/main/java/momento/sdk/exceptions/LimitExceededException.java
@@ -4,19 +4,92 @@ import momento.sdk.internal.MomentoTransportErrorDetails;
 
 /** Requested operation couldn't be completed because system limits were hit. */
 public class LimitExceededException extends MomentoServiceException {
-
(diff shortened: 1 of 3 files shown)


momento-sdk/src/main/java/momento/sdk/exceptions/LimitExceededException.java
```diff
@@ -4,19 +4,97 @@
...
+    this.messageWrapper = messageWrapper;
+  }
+
+  /* (non-Javadoc)
+   * @see java.lang.Enum#toString()
+   */
+  @Override
+  public String toString() {
+    return messageWrapper;
   }
```
malandis, line 99: This should be in its own file (as other enums are in the project).

malandis (commented):
Nice how this one is compact! Just one suggestion re: the new enum.

malandis (approved):

Issue #398, opened 2024-10-24
[Java SDK] RetryInterceptor for streaming calls


rishtigupta:
## Issue Description:
We need to implement a retry interceptor for streaming calls in the Java SDK, based on the investigation conducted in [this ticket](https://github.com/momentohq/dev-eco-issue-tracker/issues/1028). During the investigation, it was found that the same RetryInterceptor logic used for unary calls could be applied to streaming calls as well.

However, testing this implementation against Momento Local revealed an issue. After a retry is triggered, the program hangs when performing scalar or batch operations. This indicates that the behavior of Momento Local differs from that of the actual server, potentially leading to the hang.

## Next Steps:
To resolve this issue, we need to collaborate with a backend engineer to set up a development environment (dev cell) where we can simulate a "SERVER_UNAVAILABLE" error locally. This will allow us to test the retry logic under conditions that better mimic the actual server behavior, ensuring the retry mechanism functions correctly for both scalar and batch operations.

## Action Items:
[x] Coordinate with a backend engineer to configure the dev cell for simulating the "SERVER_UNAVAILABLE" error.
[x] Re-test the RetryInterceptor logic with the new setup to verify if the issue persists.
[x] Ensure that both scalar and batch operations are handled correctly during retries without causing the program to hang.

This will help isolate whether the discrepancy between Momento Local and the actual server is contributing to the issue and enable us to refine the retry logic accordingly.

## Code References:
For details on how testing against momento-local was done, check this PR: https://github.com/momentohq/client-sdk-java/pull/394 [WIP-POC]

rishtigupta:
### Investigation Summary:

During testing on my development environment (`developer-rishti-dev`), I simulated server-side errors (5XX) to examine the behavior of both unary and streaming calls under realistic service conditions, rather than using `momento-local`. The investigation revealed the following:

1. **Proper Program Exit**: The program exits without hanging when `cacheClient.close()` is explicitly called. This indicates that certain resources remain open if `cacheClient` is not closed, suggesting that the `RetryInterceptor` is keeping resources active.
2. **Retry Success**: I successfully retried streaming calls (e.g., `getBatch`) using the existing `RetryInterceptor`, confirming its effectiveness for both unary and streaming calls.

### Conclusions:

1. **Consistency Across Environments**: `momento-local` and the actual service perform identically, confirming that there are no issues with the `momento-local` implementation.
[...]

PR #400, opened 2024-11-08
feat: use retry interceptor for streaming calls

rishtigupta:
### Overview:
- This PR updates the RetryInterceptor to handle retries for both unary and streaming calls (e.g., getBatch) without requiring any code changes to the interceptor itself. The existing logic in RetryInterceptor uses the following condition to determine if a call should be retried:

```
if (!method.getType().clientSendsOneMessage()) {
  return channel.newCall(method, callOptions);
}
```
This leverages the method.getType().clientSendsOneMessage() check, where:

```
 public final boolean clientSendsOneMessage() {
            return this == UNARY || this == SERVER_STREAMING;
        }
```
Since both unary and server-side streaming calls return true for `clientSendsOneMessage()`, the existing interceptor logic is inherently compatible with server-side streaming calls. To enable retries for `getBatch` specifically, this PR simply adds it to the list of API calls that should trigger retries.

### Enhancements:
- **Resource Cleanup:** Introduced a `cancelAttempt` call to release resources if the client decides to stop retrying. This prevents the program from hanging by ensuring that any remaining resources are cleaned up.
- **API Whitelisting:** Added getBatch to the list of retriable API calls, enabling retry behavior for streaming calls without modifications to the core interceptor logic.

## Testing
To verify the retry functionality for streaming calls, I:
- Configured a development cell to simulate real server-side behavior, including 5XX errors.
- Developed a test program (below) that attempts to retrieve cached items using `getBatch`, retrying on failure. This allowed for validating that retries work as expected and that resources are released properly upon completion.

```
package momento.sdk;

import java.time.Duration;
import java.util.HashMap;
import java.util.Map;
import java.util.UUID;

import momento.sdk.auth.CredentialProvider;
import momento.sdk.config.Configuration;
import momento.sdk.config.Configurations;
import momento.sdk.responses.cache.GetBatchResponse;
import momento.sdk.responses.cache.SetBatchResponse;

public class Program {
    String cacheName = "java-test-cache";
    CredentialProvider credentialProvider = CredentialProvider.fromEnvVar("MOMENTO_API_KEY");

    Configuration configuration = Configurations.Laptop.v1();
    CacheClient cacheClient =
            new CacheClientBuilder(credentialProvider, configuration, Duration.ofMinutes(1)).build();

    final Map<String, String> items = new HashMap<>();
    for (int i = 0; i < 10; i++) {
      items.put("key" + i, "val" + i);
    }

    // Set batch operation to populate cache (uncomment if needed)
    final SetBatchResponse setBatchResponse =
        cacheClient.setBatch(cacheName, items, Duration.ofMinutes(60)).join();
    if (setBatchResponse instanceof SetBatchResponse.Success) {
      System.out.println("Set batch successful");
    } else {
      System.out.println("Failed to set batch" + setBatchResponse.toString());
      SetBatchResponse.Error error = (SetBatchResponse.Error) setBatchResponse;
      System.out.println("Failed to set batch: " + error.getErrorCode());
    }

    while (true) {
      try {
        final GetBatchResponse getBatchResponse =
                cacheClient.getBatch(cacheName, items.keySet()).join();
        if (getBatchResponse instanceof GetBatchResponse.Success) {
          System.out.println("Get batch successful");
          Map<String, String> values =
                  ((GetBatchResponse.Success) getBatchResponse).valueMapStringString();
          for (Map.Entry<String, String> entry : values.entrySet()) {
            System.out.println("Key: " + entry.getKey() + " Value: " + entry.getValue());
          }
            // Exit the loop if all items are found
            if (values.size() == items.size()) {
                break;
            }
        } else {
          System.out.println("Failed to get batch: " + getBatchResponse.toString());
          GetBatchResponse.Error error = (GetBatchResponse.Error) getBatchResponse;
          System.out.println("Failed to get batch: " + error.getErrorCode());
        }
      } catch (Exception e) {
        System.out.println("Exception during getBatch: " + e.getMessage());
      }

      try {
        // Add a delay between each retry
        Thread.sleep(5000);
      } catch (InterruptedException e) {
        Thread.currentThread().interrupt();
        break;
      }
    }

    System.out.println("Done");
    // cacheClient.close()
  }
}
```

This setup confirmed that RetryInterceptor effectively handles retries for both unary and streaming calls, ensuring robust error handling for getBatch requests.

## Issue
https://github.com/momentohq/client-sdk-java/issues/398

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
+++ b/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
@@ -18,26 +18,27 @@ import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import momento.sdk.retry.RetryEligibilityStrategy;
 import momento.sdk.retry.RetryStrategy;
-import momento.sdk.retry.RetryingUnaryClientCall;
+import momento.sdk.retry.RetryingClientCall;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * Interceptor for retrying client calls with gRPC servers. This interceptor is responsible for
- * handling retry logic when making unary (single request, single response) gRPC calls.
+ * handling retry logic when making unary (single request, single response) and streaming gRPC
+ * calls.
  *
  * <p>A {@link ClientCall} is essentially an instance of a gRPC invoker. Every gRPC interceptor
  * expects us to return such client call(s) that it will execute in order. Each call has a "start"
  * method, which is the entry point for the call.
  *
- * <p>This retry client interceptor returns an instance of a {@link RetryingUnaryClientCall}, which
- * is a client call designed to handle retrying unary (single request, single response) operations.
- * The interceptor uses a provided {@link RetryStrategy} to determine when and how to retry failed
- * calls.
+ * <p>This retry client interceptor returns an instance of a {@link RetryingClientCall}, which is a
+ * client call designed to handle retrying unary (single request, single response) and streaming
+ * call operations. The interceptor uses a provided {@link RetryStrategy} to determine when and how
+ * to retry failed calls.
  *
  * <p>When a gRPC call is intercepted, the interceptor checks whether the method is unary (client
- * sends one message), and if so, it wraps the original {@link ClientCall} with the {@link
- * RetryingUnaryClientCall}. This custom call is responsible for handling the retry logic.
+ * sends one message) or streaming, and if so, it wraps the original {@link ClientCall} with the
+ * {@link RetryingClientCall}. This custom call is responsible for handling the retry logic.
  *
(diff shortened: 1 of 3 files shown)


anitarua (approved):
lgtm!

momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
```diff
@@ -126,12 +124,13 @@ public void onClose(Status status, Metadata trailers) {
                 // a delay not present indicates we have exhausted retries or exceeded
                 // delay or any variable the strategy author wishes to not retry anymore
                 if (!retryDelay.isPresent()) {
+                  cancelAttempt();
```
nand4011, line 127: Good catch!

nand4011 (approved):

PR #402, opened 2024-11-11
feat: handle discontinuity and heartbeat in topics and add sequence p…

rishtigupta:
…age to topics

## PR Description:
- Updates the client-protos version
- Implement TopicDiscontinuity class to handle discontinuities in topics
- Handle onHeartbeat callback function

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1062


momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -155,15 +157,24 @@ private void handleSubscriptionItem(_SubscriptionItem item) {
...
   private void handleSubscriptionDiscontinuity(_SubscriptionItem discontinuityItem) {
     logger.debug(
-        "{}, {}, {}, {}",
+        "discontinuity {}, {}, {}, {}, {}, {}",
         options.getCacheName(),
         options.getTopicName(),
         discontinuityItem.getDiscontinuity().getLastTopicSequence(),
+        discontinuityItem.getDiscontinuity().getNewTopicSequence(),
+        discontinuityItem.getDiscontinuity().getLastTopicSequence(),
         discontinuityItem.getDiscontinuity().getNewTopicSequence());
```
anitarua, line 166: I think you meant to print last topic seq, new topic seq, and new seq page?

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -175,6 +186,15 @@ private void handleSubscriptionItemMessage(_SubscriptionItem item) {
     _TopicValue topicValue = topicItem.getValue();
     options.subscriptionState.setResumeAtTopicSequenceNumber(
         (int) topicItem.getTopicSequenceNumber());
+    options.subscriptionState.setResumeAtTopicSequencePage((int) topicItem.getSequencePage());
```
anitarua, line 189: given recent discussion about these types in dotnet, is `int` the type we want here? should it be a `long` or something instead to more closely match the uint64 type in the protos?

rishtigupta: Yeah, I wanted to discuss that in mid-week checkin meeting. We were originally casting the sequence number to int, and therefore I kept casting to it. But based the latest discussions, it makes sense to case them to long

anitarua (dismissed):
lgtm! 
maybe good to get a review from @nand4011 though for any Java-specific details?

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -155,15 +157,23 @@ private void handleSubscriptionItem(_SubscriptionItem item) {
 
   private void handleSubscriptionDiscontinuity(_SubscriptionItem discontinuityItem) {
     logger.debug(
```
nand4011, line 159: If we're logging here, does it make sense for the onDiscontinuity and onHeartbeat default implementations to log?

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -173,8 +183,16 @@ private void handleSubscriptionUnknown() {
   private void handleSubscriptionItemMessage(_SubscriptionItem item) {
     _TopicItem topicItem = item.getItem();
     _TopicValue topicValue = topicItem.getValue();
-    options.subscriptionState.setResumeAtTopicSequenceNumber(
-        (int) topicItem.getTopicSequenceNumber());
+    options.subscriptionState.setResumeAtTopicSequenceNumber(topicItem.getTopicSequenceNumber());
+    options.subscriptionState.setResumeAtTopicSequencePage(topicItem.getSequencePage());
+
+    logger.debug(
```
nand4011, line 189: This seems like it could add a bunch of noise to the log if debug logs are turned on and the program is connected to a topic with a lot of traffic. I would lean toward letting the user log in the callback if they want, but I could be convinced.

nand4011 (commented):
Some logging questions, but otherwise it looks good.

momento-sdk/src/main/java/momento/sdk/ISubscriptionCallbacks.java
```diff
@@ -24,6 +25,18 @@ public interface ISubscriptionCallbacks {
...
 
+  /** Called when a discontinuity occurs during the subscription. */
+  default void onDiscontinuity(TopicDiscontinuity discontinuity) {
+    logger.info("Discontinuity occurred: {}", discontinuity);
+  }
+  ;
+
+  /** Called when a heartbeat is received during the subscription. */
+  default void onHeartbeat() {
+    logger.info("Heartbeat received");
```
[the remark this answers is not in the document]
malandis: For the defaults, we should align on functionality cross-SDK. The JavaScript SDK [has void (empty) functions for this](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/internal/clients/pubsub/AbstractPubsubClient.ts#L135-L154). I'd just go with that for consistency.

momento-sdk/src/main/java/momento/sdk/SendSubscribeOptions.java
```diff
@@ -10,6 +11,8 @@ class SendSubscribeOptions implements ISubscriptionCallbacks {
   ItemCallback onItem;
   CompletedCallback onCompleted;
   ErrorCallback onError;
+  DiscontinuityCallback onDiscontinuity;
+  HeartbeatCallback onHeartbeat;
```
malandis, line 15: Any reason why these are package private vs class private?

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
malandis: Agreed. Either that or make these trace level.

momento-sdk/src/main/java/momento/sdk/responses/topic/TopicDiscontinuity.java
```diff
@@ -0,0 +1,46 @@
...
+  public String toString() {
+    return "TopicDiscontinuity{"
+        + "lastSequenceNumber="
+        + lastSequenceNumber
+        + ", newSequenceNumber="
+        + newSequenceNumber
+        + ", newSequencePage="
+        + newSequencePage
+        + '}';
+  }
```
malandis, line 24: A string builder here would save some compute, eg [here](https://github.com/momentohq/client-sdk-java/blob/main/momento-sdk/src/main/java/momento/sdk/responses/storage/StorageValue.java#L117-L133)

momento-sdk/src/intTest/java/momento/sdk/TopicClientTest.java
```diff
@@ -61,6 +62,16 @@ public void onCompleted() {
...
+
+      @Override
+      public void onDiscontinuity(TopicDiscontinuity discontinuity) {
+        logger.info("onDiscontinuity Invoked");
+      }
+
+      @Override
+      public void onHeartbeat() {
+        logger.info("onHeartbeat Invoked");
+      }
```
malandis, line 74: This could make the test logs pretty chatty. I'd either comment these out or make them debug/trace level.

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
malandis: I don't think so. We should remove logging from the default implementations and use trace logging here, following the [JS SDK conventions](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/pubsub-client.ts#L248-L264)

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -155,15 +157,23 @@ private void handleSubscriptionItem(_SubscriptionItem item) {
...
-        discontinuityItem.getDiscontinuity().getNewTopicSequence());
+        discontinuityItem.getDiscontinuity().getNewTopicSequence(),
+        discontinuityItem.getDiscontinuity().getNewSequencePage());
+
+    options.onDiscontinuity(
+        new TopicDiscontinuity(
+            discontinuityItem.getDiscontinuity().getLastTopicSequence(),
+            discontinuityItem.getDiscontinuity().getNewTopicSequence(),
+            discontinuityItem.getDiscontinuity().getNewSequencePage()));
   }
```
malandis, line 172: On a discontinuity we also have to set the subscriptionState's topic sequence number and topic sequence page to the new values from the discontinuity.

malandis (commented):
Ensure setting the subscription state internal sequence number and sequence page on a discontinuity.

I also recommend changing the topic internal log level to trace, and leaving logging out of the default callback implementations.

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
rishtigupta: Good catch! I think we missed that in JS SDK: https://github.com/momentohq/client-sdk-javascript/blob/5ead1273b4ab107cb2343ff302ec1fc053acbe50/packages/client-sdk-web/src/internal/pubsub-client.ts#L229-L239?

momento-sdk/src/main/java/momento/sdk/SendSubscribeOptions.java
rishtigupta: That might have been an oversight when we implemented topics in java sdk. 
I don't see other classes accessing these callback functions.

rishtigupta: We can maybe address this in a separate PR.

malandis (approved):

PR #408, opened 2024-12-11
feat: add the read concern header

nand4011:
Add the read concern option to the cache Configuration and thread it through to the gRPC internals of the client.

Add an argument to the UserHeaderInterceptor constructor representing a list of extra headers to add to every request. The stubs manager uses this to add the read concern header if it isn't set to balanced.

Make the base cache test class create clients with both balanced and consistent read concern, and set the cache client used by all the tests to the consistent one if the CONSISTENT_READS env var is defined.

Add tasks to run integration tests for the different clients.

Add make targets to use the new tasks.

### DIFF
--- a/momento-sdk/build.gradle.kts
+++ b/momento-sdk/build.gradle.kts
@@ -82,18 +82,12 @@ fun registerIntegrationTestTask(name: String, testClasses: List<String>) {
 
 registerIntegrationTestTask(
     "test-auth-service",
-    listOf("momento.sdk.AuthClient*")
+    listOf("momento.sdk.auth.*")
 )
 
 registerIntegrationTestTask(
     "test-cache-service",
-    listOf(
-        "momento.sdk.Cache*",
-        "momento.sdk.DictionaryTest",
-        "momento.sdk.ListTest",
-        "momento.sdk.SetTest",
-        "momento.sdk.SortedSetTest"
-    )
+    listOf("momento.sdk.cache.*")
 )
 
 registerIntegrationTestTask(
@@ -103,5 +97,5 @@ registerIntegrationTestTask(
 
 registerIntegrationTestTask(
     "test-topics-service",
-    listOf("momento.sdk.Topic*")
+    listOf("momento.sdk.topics.*")
 )
--- a/momento-sdk/build.gradle.kts
+++ b/momento-sdk/build.gradle.kts
@@ -1,3 +1,5 @@
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+
 plugins {
     id("momento.publishable-java-lib")
     id("momento.junit-tests")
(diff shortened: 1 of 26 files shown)


momento-sdk/src/intTest/java/momento/sdk/BaseCacheTestClass.java
```diff
@@ -15,22 +17,34 @@ public class BaseCacheTestClass {
...
-            .build();
+
+    final Configuration config = Configurations.Laptop.latest();
+    final Configuration consistentConfig = config.withReadConcern(ReadConcern.CONSISTENT);
+
+    balancedReadCacheClient =
+        CacheClient.builder(credentialProvider, config, DEFAULT_TTL_SECONDS).build();
+    consistentReadCacheClient =
+        CacheClient.builder(credentialProvider, consistentConfig, DEFAULT_TTL_SECONDS).build();
+    cacheClient = consistentReads ? consistentReadCacheClient : balancedReadCacheClient;
```
pgautier404, line 37: Is there any reason to instantiate both clients here? Not a huge deal, but if you're only using one of the two it seems like you could just use the env var to determine which config to use and only ever instantiate a single cacheClient.

momento-sdk/build.gradle.kts
```diff
@@ -53,3 +55,53 @@ tasks.named("analyzeIntTestClassesDependencies").configure {
...
+
+registerIntegrationTestTask(
+    "test-storage-service",
+    listOf("momento.sdk.storage.*")
+)
+
+registerIntegrationTestTask(
+    "test-topics-service",
+    listOf("momento.sdk.Topic*")
+)
```
malandis, line 107: Can we group each service under a package of the same name? That was a change I made in other SDKs to make the pattern matching more maintainable, eg:

- move all auth related tests under a package `momento.sdk.auth`
- cache under `momento.sdk.cache`
- etc

momento-sdk/src/intTest/java/momento/sdk/BaseCacheTestClass.java
```diff
@@ -52,8 +66,4 @@ public static void cleanupTestCache(String cacheName) {
   public static String testCacheName() {
     return "java-integration-test-default-" + UUID.randomUUID();
   }
-
-  public static String testStoreName() {
-    return "java-integration-test-default-" + UUID.randomUUID();
-  }
```
malandis, line 58: Nice find

momento-sdk/src/intTest/java/momento/sdk/CacheControlPlaneTest.java
```diff
@@ -199,17 +199,4 @@ public void throwsInvalidArgumentForNegativeRequestTimeout() {
...
-    //noinspection resource
-    assertThatExceptionOfType(InvalidArgumentException.class)
-        .isThrownBy(
-            () ->
-                CacheClient.builder(
-                        credentialProvider,
-                        Configurations.Laptop.latest().withTimeout(null),
-                        DEFAULT_TTL_SECONDS)
-                    .build());
-  }
```
malandis, line 214: Was this no longer needed?

momento-sdk/src/main/java/momento/sdk/ScsDataGrpcStubsManager.java
```diff
@@ -71,7 +75,7 @@ final class ScsDataGrpcStubsManager implements AutoCloseable {
 
   private final ExecutorService retryExecutor;
 
-  // An arbitary selection of twice the number of available processors.
+  // An arbitrary selection of twice the number of available processors.
```
malandis, line 78: 🦅 👁️

momento-sdk/src/main/java/momento/sdk/config/ReadConcern.java
```diff
@@ -0,0 +1,19 @@
...
+   * Balanced read concern makes no consistency guarantee. The default read concern for the cache
+   * client.
+   */
+  BALANCED,
+  /** Consistent read concern guarantees read after write consistency. */
+  CONSISTENT;
+
+  public String toLowerCase() {
+    return name().toLowerCase();
+  }
```
malandis, line 18: 👍

malandis (commented):
LGTM just a couple minor questions

momento-sdk/src/intTest/java/momento/sdk/BaseCacheTestClass.java
nand4011: That was done in other SDKs because we had tests that required the specific clients. I was on the fence about adding tests that explicitly check that the client works with each read concern, since they run with the read concern header on github and the canaries, but run without it locally. If I add those tests, we need the clients. Otherwise I'll remove them.

malandis (approved):

PR #409, opened 2025-01-09
chore: Add a gRPC configuration interface

nand4011:
Add an interface for GrpcConfigration so that common code for setting up connections can work with the different implementations.

Update the applyGrpcConfigurationToChannelBuilder logic to work with IGrpcConfigration and reduce the two implementations to one.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/config/transport/GrpcConfiguration.java
+++ b/momento-sdk/src/main/java/momento/sdk/config/transport/GrpcConfiguration.java
@@ -5,17 +5,18 @@ import static momento.sdk.ValidationUtils.ensureRequestDeadlineValid;
 import java.time.Duration;
 import java.util.Optional;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import momento.sdk.internal.GrpcChannelOptions;
 
 /** Abstracts away the gRPC configuration tunables. */
 public class GrpcConfiguration implements IGrpcConfiguration {
 
-  private final Duration deadline;
+  private final @Nonnull Duration deadline;
   private final int minNumGrpcChannels;
-  private final Optional<Integer> maxMessageSize;
-  private final Optional<Boolean> keepAliveWithoutCalls;
-  private final Optional<Integer> keepAliveTimeoutMs;
-  private final Optional<Integer> keepAliveTimeMs;
+  private final @Nullable Integer maxMessageSize;
+  private final @Nullable Boolean keepAliveWithoutCalls;
+  private final @Nullable Duration keepAliveTimeout;
+  private final @Nullable Duration keepAliveTime;
 
   /**
    * Constructs a GrpcConfiguration.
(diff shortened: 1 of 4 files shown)


momento-sdk/src/main/java/momento/sdk/config/transport/IGrpcConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+   * @return the time to wait for a keepalive ping response before considering the connection dead.
+   */
+  Optional<Integer> getKeepAliveTimeoutMs();
+
+  /**
+   * The time to wait between keepalive pings.
+   *
+   * @return the time to wait between keepalive pings.
+   */
+  Optional<Integer> getKeepAliveTimeMs();
```
malandis, line 48: Maybe make these `Optional<Duration>` for ease of use internally?

momento-sdk/src/main/java/momento/sdk/config/transport/IGrpcConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * The minimum number of gRPC channels to keep open at any given time.
+   *
+   * @return the minimum number of gRPC channels.
+   */
+  int getMinNumGrpcChannels();
+
+  /**
+   * The maximum size of a message (in bytes) that can be received by the client.
```
malandis, line 23: Does this apply for received only or sent and received? If the former then let's adjust the name.

momento-sdk/src/main/java/momento/sdk/config/transport/IGrpcConfiguration.java
nand4011: The channel builder doesn't have methods that take Duration, so we would need to convert it back to an int anyway.

momento-sdk/src/main/java/momento/sdk/config/transport/IGrpcConfiguration.java
nand4011: These methods all already exist and are publicly accessible, so we can't change them. I could add new methods and leave the old ones out of the interface if we want to change the names or return types.
The channel builder doesn't have a way to specify the max outbound message size from what I can see, so this is receive only.

momento-sdk/src/main/java/momento/sdk/config/transport/IGrpcConfiguration.java
malandis: Right -- but for the user or us (internally) it would be easier to set the values, right?

momento-sdk/src/main/java/momento/sdk/config/transport/GrpcConfiguration.java
```diff
@@ -51,13 +52,41 @@ public GrpcConfiguration(
       Optional<Boolean> keepAliveWithoutCalls,
       Optional<Integer> keepAliveTimeout,
       Optional<Integer> keepAliveTime) {
+    this(
+        deadline,
+        minNumGrpcChannels,
+        maxMessageSize.orElse(null),
+        keepAliveWithoutCalls.orElse(null),
+        keepAliveTimeout.map(Duration::ofMillis).orElse(null),
```
malandis, line 60: Nice

malandis (approved):

PR #410, opened 2025-01-10
feat: Add leaderboard upsert, fetch by rank, and remove elements

nand4011:
Add a new client for interacting with Momento leaderboards and the configuration objects required for creating one.

Add the leaderboard upsert, fetch by rank, and remove elements methods.

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/leaderboard/LeaderboardClientTest.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/leaderboard/LeaderboardClientTest.java
@@ -1,0 +1,327 @@
+package momento.sdk.leaderboard;
+
+import static momento.sdk.TestUtils.randomString;
+import static org.assertj.core.api.Assertions.*;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import momento.sdk.ILeaderboard;
+import momento.sdk.exceptions.CacheNotFoundException;
+import momento.sdk.exceptions.InvalidArgumentException;
+import momento.sdk.responses.SortOrder;
+import momento.sdk.responses.leaderboard.FetchResponse;
+import momento.sdk.responses.leaderboard.LeaderboardElement;
+import momento.sdk.responses.leaderboard.RemoveElementsResponse;
+import momento.sdk.responses.leaderboard.UpsertResponse;
+import org.assertj.core.api.InstanceOfAssertFactories;
+import org.junit.jupiter.api.Test;
+
+public class LeaderboardClientTest extends BaseLeaderboardTestClass {
+
+  // upsert
+
+  @Test
+  public void upsertHappyPath() {
+    final String leaderboardName = randomString("leaderboard");
+    final ILeaderboard leaderboard = leaderboardClient.leaderboard(cacheName, leaderboardName);
+
+    final Map<Integer, Double> elements = new HashMap<>();
+    elements.put(1, 1.0);
+    elements.put(2, 2.0);
+    elements.put(3, 3.0);
+
+    assertThat(leaderboard.upsert(elements))
+        .succeedsWithin(FIVE_SECONDS)
+        .isInstanceOf(UpsertResponse.Success.class);
+
+    assertThat(leaderboard.fetchByRank(0, 10, SortOrder.ASCENDING))
+        .succeedsWithin(FIVE_SECONDS)
+        .asInstanceOf(InstanceOfAssertFactories.type(FetchResponse.Success.class))
+        .satisfies(
+            resp -> {
+              final List<LeaderboardElement> scoredElements = resp.elementsList();
+              assertThat(scoredElements).hasSize(3);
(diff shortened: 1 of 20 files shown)


momento-sdk/src/main/java/momento/sdk/responses/leaderboard/LeaderboardElement.java
```diff
@@ -0,0 +1,66 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * A pairing of an integer ID to a score. LeaderboardElements have equality based on their ID and
+ * ordering based on their score.
+ */
+public class LeaderboardElement implements Comparable<LeaderboardElement> {
+  private final int id;
+  private final double score;
```
anitarua, line 11: is this also missing rank?

momento-sdk/src/intTest/java/momento/sdk/leaderboard/LeaderboardClientTest.java
```diff
@@ -0,0 +1,327 @@
...
+import org.junit.jupiter.api.Test;
+
+public class LeaderboardClientTest extends BaseLeaderboardTestClass {
+
+  // upsert
+
+  @Test
+  public void upsertHappyPath() {
+    final String leaderboardName = randomString("leaderboard");
+    final ILeaderboard leaderboard = leaderboardClient.leaderboard(cacheName, leaderboardName);
```
anitarua, line 29: tests should make sure to remove elements from leaderboards / delete leaderboards since leaderboards have 7-day ttl by default

malandis (commented):
Few questions about the UX which come from the leaderboard service. Am I correct that we don't explicitly create a leaderboard resource? Ie we have no way to "list leaderboards" or "delete a leaderboard"?

momento-sdk/src/intTest/java/momento/sdk/leaderboard/LeaderboardClientTest.java
malandis: Is that necessary if we delete the test cache at the end of the suite?

anitarua:
> Few questions about the UX which come from the leaderboard service. Am I correct that we don't explicitly create a leaderboard resource? Ie we have no way to "list leaderboards" or "delete a leaderboard"?

There's a [delete leaderboard API](https://docs.momentohq.com/leaderboards/api-reference#delete-leaderboard). 
No list leaderboards, but I don't see a list [sorted sets](https://docs.momentohq.com/cache/develop/api-reference/sorted-set-collections) in the regular cache methods either.

momento-sdk/src/main/java/momento/sdk/responses/leaderboard/LeaderboardElement.java
nand4011: Good call. I created this from ScoredElement, which doesn't have rank and missed it in the JS version.

momento-sdk/src/intTest/java/momento/sdk/leaderboard/LeaderboardClientTest.java
anitarua: good point, deleting the cache also works

momento-sdk/src/main/java/momento/sdk/LeaderboardClient.java
```diff
@@ -0,0 +1,44 @@
...
+   */
+  public static LeaderboardClientBuilder builder(
+      CredentialProvider credentialProvider, LeaderboardConfiguration configuration) {
+    return new LeaderboardClientBuilder(credentialProvider, configuration);
+  }
+
+  /** Creates a leaderboard instance with the given cache and leaderboard names. */
+  public ILeaderboard leaderboard(@Nonnull String cacheName, @Nonnull String leaderboardName) {
+    return new Leaderboard(this.dataClient, cacheName, leaderboardName);
+  }
```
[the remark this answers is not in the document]
anitarua: I think the service only creates the leaderboard after elements are upserted, but it will return no-op / length 0 / empty list if you perform any operations on it before upserting anything

momento-sdk/src/main/java/momento/sdk/LeaderboardGrpcStubsManager.java
```diff
@@ -0,0 +1,84 @@
...
+    final NettyChannelBuilder channelBuilder =
+        NettyChannelBuilder.forAddress(credentialProvider.getCacheEndpoint(), 443);
+
+    // set additional channel options (message size, keepalive, auth, etc)
+    GrpcChannelOptions.applyGrpcConfigurationToChannelBuilder(
+        configuration.getTransportStrategy().getGrpcConfiguration(), channelBuilder);
+
+    final List<ClientInterceptor> clientInterceptors = new ArrayList<>();
+    clientInterceptors.add(
+        new UserHeaderInterceptor(credentialProvider.getAuthToken(), "leaderboard"));
```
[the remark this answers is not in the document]
nand4011: The UserHeaderInterceptor already handles that.

anitarua (approved):
lgtm!

PR #411, opened 2025-01-10
Add a basic leaderboard example that uses a SNAPSHOT version of the sdk

nand4011:


nand4011:
I updated the pr to move the leaderboard examples into their own module and to use the released SDK version with leaderboard support.

nand4011:
Yep. Good call.

PR #412, opened 2025-01-15
feat: add the remaining leaderboard APIs

nand4011:
Add leaderboard fetchByRank, getRank, length, and delete APIs.

Set the default leaderboard max received message size to 200MB.

Make SortOrder in fetchByRank nullable and have it default to ascending if null.

Make the different null check validation methods use the same backing validateNotNull.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/ILeaderboard.java
+++ b/momento-sdk/src/main/java/momento/sdk/ILeaderboard.java
@@ -3,8 +3,11 @@ package momento.sdk;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import momento.sdk.responses.SortOrder;
+import momento.sdk.responses.leaderboard.DeleteResponse;
 import momento.sdk.responses.leaderboard.FetchResponse;
+import momento.sdk.responses.leaderboard.LengthResponse;
 import momento.sdk.responses.leaderboard.RemoveElementsResponse;
 import momento.sdk.responses.leaderboard.UpsertResponse;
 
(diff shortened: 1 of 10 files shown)


momento-sdk/src/main/java/momento/sdk/ILeaderboard.java
```diff
@@ -31,20 +73,48 @@ public interface ILeaderboard {
...
   CompletableFuture<FetchResponse> fetchByRank(
-      int startRank, int endRank, @Nonnull SortOrder order);
+      int startRank, int endRank, @Nullable SortOrder order);
+
+  /**
+   * Looks up the rank of the given elements in the leaderboard. The returned elements will be
+   * sorted numerically by their IDs. Note: rank is 0-based (index begins at 0).
+   *
+   * @param ids The IDs of the elements to fetch from the leaderboard.
+   * @param order The order to fetch the elements in. Defaults to {@link SortOrder#ASCENDING}.
```
anitarua, line 88: maybe good to specify `order` as "The rank order to fetch..." given our previous confusion of returned order of elements?

momento-sdk/src/main/java/momento/sdk/Leaderboard.java
```diff
@@ -25,14 +28,48 @@ public CompletableFuture<UpsertResponse> upsert(@Nonnull Map<Integer, Double> el
...
+      @Nullable Double maxScore,
+      @Nullable SortOrder order,
+      @Nullable Integer offset,
+      @Nullable Integer count) {
+    return leaderboardDataClient.fetchByScore(
+        cacheName, leaderboardName, minScore, maxScore, order, offset, count);
+  }
+
+  @Override
+  public CompletableFuture<FetchResponse> fetchByScore(
```
anitarua, line 43: it might be nice to have a `fetchByScore` that accepts no arguments at all since they're all optional/nullable parameters

anitarua (commented):
lgtm overall!

momento-sdk/src/main/java/momento/sdk/Leaderboard.java
nand4011: We do have an equivalent java sorted set method like that. I'll add one in.

anitarua (approved):
lgtm!

PR #413, opened 2025-01-15
fix: update a comment that fails the build in some instances

nand4011:
Remove a non-ASCII character from a javadoc comment that was preventing the canary docker image from building.


malandis (approved):
That's weird. 😨

rishtigupta (approved):
:shipit:

PR #414, opened 2025-01-16
chore: consolidate gRPC stubs handling logic into an abstract class

nand4011:
Create an abstract gRPC stubs manager that the existing stubs managers can descend from. Almost all the logic in the stubs managers is shared, so it should go into one place. Retries and the read concern header are moved into the common logic, even though they are only used in the cache data manager, to make them easier to use by other managers in the future.

Make the close method in the stubs managers await termination of the gRPC channels and retry executors. This ensures they will try to complete any in-flight calls before shutting down.

Propagate the thread interruption flag in the base internal client close method to make it work better with other close methods.

Add withDeadline to the grpc configuration interface so it can be used in the stubs managers.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/StorageDataGrpcStubsManager.java
+++ b/momento-sdk/src/main/java/momento/sdk/StorageDataGrpcStubsManager.java
@@ -1,28 +1,9 @@
 package momento.sdk;
 
 import grpc.store.StoreGrpc;
-import io.grpc.ClientInterceptor;
-import io.grpc.ConnectivityState;
-import io.grpc.ManagedChannel;
-import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder;
-import java.time.Duration;
-import java.time.Instant;
-import java.util.ArrayList;
-import java.util.List;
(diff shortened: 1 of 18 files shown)


momento-sdk/src/main/java/momento/sdk/StorageDataGrpcStubsManager.java
```diff
@@ -31,141 +12,18 @@
...
-  public void close() {
-    for (ManagedChannel channel : channels) {
-      channel.shutdown();
-    }
+      @Nonnull CredentialProvider credentialProvider,
+      @Nonnull IGrpcConfiguration grpcConfiguration) {
+    super(
+        AbstractGrpcStubsManager.Config.create(
+            "store",
+            credentialProvider.getCacheEndpoint(),
```
nand4011, line 24: Should this use `getStorageEndpoint()`? It was using `getCacheEndpoint()` before, but the credential provider has a method for using the storage endpoint.

anitarua: yeah, probably should be using storage endpoint

PR #417, opened 2025-01-21
feat: implement momento local provider for momento local

rishtigupta:
## PR Description:
- Implement `MomentoLocalProvider`for momento local. 

## Testing:

<details close>
  <summary>Tested using the following Java program:</summary>
  
```
public class Program {
    String cacheName = "test-cache-java";
    CredentialProvider credentialProvider = new MomentoLocalProvider("0.0.0.0", 8080);

    Configuration configuration = Configurations.Laptop.latest();
    CacheClient cacheClient =
        new CacheClientBuilder(credentialProvider, configuration, Duration.ofMinutes(1)).build();

    CacheCreateResponse createResponse = cacheClient.createCache(cacheName).join();
    if (createResponse instanceof CacheCreateResponse.Success) {
      System.out.println("Cache created successfully");
    } else {
      System.out.println("Failed to create cache" + createResponse.toString());
      CacheCreateResponse.Error error = (CacheCreateResponse.Error) createResponse;
      System.out.println("Failed to create cache: " + error.getErrorCode());
    }

    SetResponse setResponse = cacheClient.set(cacheName, "key", "value").join();
    if (setResponse instanceof SetResponse.Success) {
      System.out.println("Set key successfully");
    } else {
      System.out.println("Failed to set key" + setResponse.toString());
      SetResponse.Error error = (SetResponse.Error) setResponse;
      System.out.println("Failed to set key: " + error.getErrorCode());
      cacheClient.close();
    }

    GetResponse getResponse = cacheClient.get(cacheName, "key").join();
    if (getResponse instanceof GetResponse.Hit) {
      GetResponse.Hit hit = (GetResponse.Hit) getResponse;
      System.out.println("Got value: " + hit.value());
    } else {
      System.out.println("Failed to get key" + getResponse.toString());
    }
  }
}
```

</details>

**Result from The Java test:**
```
> Task :momento-sdk:momento.sdk.Program.main()
Cache created successfully
Set key successfully
Got value: value
```

**Result from momento-local:**
```
[2025-01-21T20:34:36Z INFO  momento_local] Server listening on 0.0.0.0:8080
[2025-01-21T20:34:43Z INFO  momento_local::cache::service::local_cache_service] Processing Set request for cache: test-cache-java and key: [107, 101, 121]
[2025-01-21T20:34:43Z INFO  momento_local::cache::service::local_cache_service] Set request processed successfully for cache: test-cache-java
[2025-01-21T20:34:43Z INFO  momento_local::cache::service::local_cache_service] Processing Get request for cache: test-cache-java and key: [107, 101, 121]
[2025-01-21T20:34:43Z DEBUG momento_local::cache::service::local_cache_service] Cache hit for cache: test-cache-java and key: [107, 101, 121]
```

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1104

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
+++ b/momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
@@ -75,12 +75,12 @@ public class MomentoLocalProvider extends CredentialProvider {
     return isSecureConnection(storageEndpoint);
   }
 
-  private boolean isSecureConnection(String endpoint) {
-    return endpoint.startsWith("https://");
-  }
-
-  /** Returns the port used by this credential provider. */
+  @Override
   public int getPort() {
     return port;
+  }
+
+  private boolean isSecureConnection(String endpoint) {
+    return endpoint.startsWith("https://");
   }
 }
--- a/momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
+++ b/momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
@@ -56,8 +56,8 @@ public class MomentoLocalProvider extends CredentialProvider {
   }
 
   @Override
-  public boolean isEndpointSecure() {
-    return isSecureConnection(DEFAULT_HOSTNAME);
+  public boolean isEndpointSecure(String endpoint) {
+    return isSecureConnection(endpoint);
   }
 
   @Override
--- a/momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
+++ b/momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
@@ -18,6 +18,14 @@ public class MomentoLocalProvider extends CredentialProvider {
     this.port = port;
   }
 
+  public MomentoLocalProvider(String hostname) {
+    this(hostname, DEFAULT_PORT);
+  }
+
+  public MomentoLocalProvider(int port) {
+    this(DEFAULT_HOSTNAME, port);
+  }
+
   public MomentoLocalProvider() {
     this(DEFAULT_HOSTNAME, DEFAULT_PORT);
   }
--- a/momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
+++ b/momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
@@ -56,16 +56,12 @@ public class MomentoLocalProvider extends CredentialProvider {
   }
 
   @Override
-  public boolean isEndpointSecure(String endpoint) {
-    return isSecureConnection(endpoint);
+  public boolean isEndpointSecure() {
+    return false;
   }
 
   @Override
   public int getPort() {
     return port;
-  }
-
-  private boolean isSecureConnection(String endpoint) {
-    return endpoint.startsWith("https://");
   }
 }
--- a/momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
+++ b/momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
@@ -1,0 +1,80 @@
+package momento.sdk.auth;
+
+public class MomentoLocalProvider extends CredentialProvider {
+    private final String cacheEndpoint;
+    private final String controlEndpoint;
+    private final String tokenEndpoint;
+    private final String storageEndpoint;
+    private final int port;
+
+    private static final String DEFAULT_HOSTNAME = "127.0.0.1";
+    private static final int DEFAULT_PORT = 8080;
+
+    public MomentoLocalProvider(String hostname, int port) {
+        this.cacheEndpoint = hostname;
+        this.controlEndpoint = hostname;
+        this.tokenEndpoint = hostname;
+        this.storageEndpoint = hostname;
+        this.port = port;
+    }
+
+    public MomentoLocalProvider() {
+        this(DEFAULT_HOSTNAME, DEFAULT_PORT);
+    }
+
+    @Override
+    public String getAuthToken() {
+        return "";
+    }
+
+    @Override
+    public String getCacheEndpoint() {
+        return cacheEndpoint;
+    }
+
+    @Override
+    public boolean isCacheEndpointSecure() {
+        return isSecureConnection(cacheEndpoint);
+    }
+
+    @Override
+    public String getControlEndpoint() {
+        return controlEndpoint;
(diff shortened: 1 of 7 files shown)


momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
```diff
@@ -0,0 +1,86 @@
...
+  public String getStorageEndpoint() {
+    return storageEndpoint;
+  }
+
+  @Override
+  public boolean isStorageEndpointSecure() {
+    return isSecureConnection(storageEndpoint);
+  }
+
+  private boolean isSecureConnection(String endpoint) {
```
nand4011, line 78: Why do we need separate calls to check if each type of endpoint is secure? Will there be a credential provider that has some tls enabled endpoints and some that are not?

momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
```diff
@@ -0,0 +1,86 @@
...
+    return endpoint.startsWith("https://");
+  }
+
+  /** Returns the port used by this credential provider. */
+  public int getPort() {
```
nand4011, line 83: Should this be added to abstract credential provider? Then we could always call getPort() when setting up a channel, and we would only need the isSecure call for the channel options.

momento-sdk/src/main/java/momento/sdk/internal/GrpcChannelOptions.java
```diff
@@ -24,8 +24,18 @@ public class GrpcChannelOptions {
...
+  }
+
+  public static void applyGrpcConfigurationToChannelBuilder(
+      IGrpcConfiguration grpcConfig, NettyChannelBuilder channelBuilder, boolean isSecure) {
+    if (isSecure) {
+      channelBuilder.useTransportSecurity();
+      channelBuilder.disableRetry();
+    } else {
+      channelBuilder.usePlaintext();
+      channelBuilder.enableRetry();
```
nand4011, line 37: Why do we enable retries for non-secure endpoints?

momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
nand4011: A test that needed both a local and non-local client could also use separate credential providers.

rishtigupta: I copied this configuration from javascript sdk where each endpoint is being tested for security. 
I will work on some uit/integ tests for this

nand4011: @anitarua do you remember why we have separate methods in the JS SDK? I'm not opposed to them if we need them, but it seems like extra work to add them if they all return the same hardcoded true or false.

momento-sdk/src/main/java/momento/sdk/internal/GrpcChannelOptions.java
anitarua: momento-local is not secure, and we'll need it to write automated retry tests

momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
anitarua: I'd implemented them as separate methods to use them as part of the ternary statements when creating the grpc connection (examples: [internal data client](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/cache-data-client.ts#L217-L219), [internal control client](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/cache-control-client.ts#L69-L71)).

And they were separate methods because I'd implemented the [Endpoint](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/internal/utils/auth.ts#L38-L41) class to contain both the endpoint and whether it should be a secure connection because I thought it was possible to override individual endpoints.
But upon second look, I think the only overrides available are base endpoint or all endpoints, so this (and the JS and Go) implementations can be simplified, actually

anitarua: I vote for the simpler/more maintainable implementation here. Made a backlog ticket for looking into refactoring the other ones at a later time: https://github.com/momentohq/dev-eco-issue-tracker/issues/1111

momento-sdk/src/main/java/momento/sdk/internal/GrpcChannelOptions.java
anitarua: @rishtigupta I think you'll want to add retries back in here

rishtigupta: I tried running a momento-local instance without this and it worked, but you're right on needing it for the automated tests. Will add that in the next pr revision.

nand4011: But why do we need grpc channel level retries? Don't we disable those intentionally and use the retry layer we built?

anitarua: oh, I didn't realize that was for a grpc-level retry layer, so then yeah I guess it wouldn't be needed

momento-sdk/src/main/java/momento/sdk/auth/MomentoLocalProvider.java
```diff
@@ -0,0 +1,71 @@
...
+  public boolean isEndpointSecure(String endpoint) {
+    return isSecureConnection(endpoint);
+  }
+
+  @Override
+  public int getPort() {
+    return port;
+  }
+
+  private boolean isSecureConnection(String endpoint) {
```
anitarua, line 68: I might've missed this somewhere, but is it possible to override individual endpoints in CredentialProvider in java sdk?
If not, maybe we can remove the `String endpoint` parameter and have the `isSecureConnection` apply to all endpoints? (i.e. _all_ endpoints in a CredentialProvider are secure or insecure)

rishtigupta: I don't see a way to override individual endpoints in java sdk as of now. I can make the function return false for all endpoints for local provider though

anitarua (approved):
lgtm!

nand4011 (approved):
Looks good!

PR #418, opened 2025-01-28
feat: add test retry metrics collector

rishtigupta:
## PR Description:
- Add MomentoRPCMethod enum
- Add TestRetryMetricsCollector that would be used to collect the timestamps and retry count at the time of retries
- Add unit tests for TestRetryMetricsCollector class. 

## Issue
https://github.com/momentohq/client-sdk-java/issues/419

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/MomentoRpcMethod.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/MomentoRpcMethod.java
@@ -1,4 +1,4 @@
-package momento.sdk.retry;
+package momento.sdk;
 
 public enum MomentoRpcMethod {
   GET("_GetRequest"),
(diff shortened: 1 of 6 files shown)


anitarua (commented):
lgtm overall!
just wondering if all of these files are supposed to be in the test/java/momento/sdk/retry folder though? 
I thought js sdk defined all these in test folder

rishtigupta:
> lgtm overall! just wondering if all of these files are supposed to be in the test/java/momento/sdk/retry folder though? I thought js sdk defined all these in test folder

The problem in java sdk is that unit and integ tests are spread across different folders and thus to use the test collector in both the folders, I thought it would be better to place the collector inside the main directory where both int and unit tests can consume it. wdyt?

anitarua (approved):
oh gotcha, well I think that's fine, lgtm!

probably good to get second opinion from @nand4011 too though in case I overlooked any Java-specific concerns

rishtigupta:
Note:
In the latest revision (04b2d1724d68a84912d3c6b1185bbabab407d8cb), I have moved the TestMetricsCollector to the integ test dir, while keeping the MOmentoRPCMethod enum to the main dir since that would be made publically accessible to users once we add the MomentoLocalTestConfigMiddleware later.

nand4011 (approved):

anitarua (approved):

PR #427, opened 2025-02-11
feat: add automated retry tests to ci

rishtigupta:
## PR Description:
- Add automated retry tests to ci job

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1113

### DIFF
--- a/.github/workflows/retry-tests.yml
+++ b/.github/workflows/retry-tests.yml
@@ -7,6 +7,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      MOMENTO_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
 
     steps:
       - name: Checkout project
--- a/.github/workflows/retry-tests.yml
+++ b/.github/workflows/retry-tests.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Start Momento Local
         run: |
-          docker-compose -f docker-compose.yml up -d
+          docker run -d -p 8080:8080 gomomento/momento-local
 
       - name: Build and test with Momento Local
         env:
@@ -37,4 +37,4 @@ jobs:
       - name: Clean up Docker Container
         if: always()
         run: |
-          docker-compose -f docker-compose.yml down
+          docker rm -f momento-local || true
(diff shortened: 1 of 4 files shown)


.github/workflows/retry-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+          distribution: 'corretto'
+
+      - name: Build project
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: clean build
+
+      - name: Start Momento Local
+        run: |
+          docker run -d -p 8080:8080 gomomento/momento-local
```
nand4011, line 30: Is it worth adding a docker compose file to the repo to make it easier to run momento local here and locally?

Makefile
```diff
@@ -50,6 +50,10 @@ test-topics-service:
 test-http-service:
 	@echo "No tests for http service."
 
+## Run the retry service tests
+test-retry-service:
```
nand4011, line 54: I would name this something like `test-retries` or `test-momento-local` since there is no separate retry service/client.

nand4011 (approved):

.github/workflows/retry-tests.yml
```diff
@@ -0,0 +1,40 @@
...
+          distribution: 'corretto'
+
+      - name: Build project
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: clean build
+
+      - name: Start Momento Local
+        run: |
+          docker-compose -f docker-compose.yml up -d
```
nand4011, line 30: Are you able to use `docker compose up` and `docker compose down` in an action?

rishtigupta: I was able to do in my local. But I feel, we'll need to install docker-compose in the actions here to correctly use it

rishtigupta: If the concern is running it locally vs ci, we can create a shell script that would spin up a momento-local docker container. Then, i can use that script locally as well as in the ci actions?

nand4011: Having a docker compose seems worth for local developer use. We can just use the docker commands you were using before in the action. I forgot that docker-compose had to be installed separately.

nand4011 (approved):

PR #428, opened 2025-02-12
feat: add exponential backoff retry strategy

nand4011:
Add a new retry strategy that waits for a jittered, exponentially increasing time between attempts until the deadline passes.

### DIFF
--- a/momento-sdk/src/intTest/java/momento/sdk/retry/FixedDelayRetryStrategyIntegTest.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/retry/FixedDelayRetryStrategyIntegTest.java
@@ -66,7 +66,7 @@ public class FixedDelayRetryStrategyIntegTest {
 
   @Test
   void testNonRetryEligibleApi_shouldMakeNoAttempts_WhenFullNetworkOutage() throws Exception {
-    RetryEligibilityStrategy eligibilityStrategy = (status, methodName) -> true;
+    RetryEligibilityStrategy eligibilityStrategy = (status, methodName) -> false;
 
     FixedDelayRetryStrategy retryStrategy =
         new FixedDelayRetryStrategy(maxAttempts, delayMillis, maxDelayMillis, eligibilityStrategy);
@@ -91,7 +91,7 @@ public class FixedDelayRetryStrategyIntegTest {
           assertThat(
                   testRetryMetricsCollector.getTotalRetryCount(
                       cacheName, MomentoRpcMethod.INCREMENT))
-              .isGreaterThanOrEqualTo(0);
+              .isEqualTo(0);
         });
   }
 
--- a/momento-sdk/src/intTest/java/momento/sdk/retry/FixedDelayRetryStrategyIntegTest.java
+++ b/momento-sdk/src/intTest/java/momento/sdk/retry/FixedDelayRetryStrategyIntegTest.java
@@ -96,7 +96,7 @@ public class FixedDelayRetryStrategyIntegTest {
   }
 
   @Test
-  void testNonRetryEligibleApi_shouldMakeLessThanMaxAttempts_WhenTemporaryNetworkOutage()
+  void testRetryEligibleApi_shouldMakeLessThanMaxAttempts_WhenTemporaryNetworkOutage()
       throws Exception {
     RetryEligibilityStrategy eligibilityStrategy = (status, methodName) -> true;
 
--- a/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
+++ b/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
@@ -51,8 +51,6 @@ import org.slf4j.LoggerFactory;
  *
  * @see RetryStrategy
  * @see RetryEligibilityStrategy
- * @param <ReqT> The type of the request message.
- * @param <RespT> The type of the response message.
  */
 final class RetryClientInterceptor implements ClientInterceptor {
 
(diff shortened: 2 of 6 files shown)


momento-sdk/src/main/java/momento/sdk/retry/ExponentialBackoffRetryStrategy.java
```diff
@@ -0,0 +1,66 @@
...
+
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import java.time.Duration;
+import java.util.Optional;
+
+public class ExponentialBackoffRetryStrategy implements RetryStrategy {
+  private static final int GROWTH_FACTOR = 2;
+
+  private final long initialDelayMillis;
```
rishtigupta, line 11: Can we set default values for these params too? 
JS sdk ref: https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/config/retry/exponential-backoff-retry-strategy.ts#L22-L32

momento-sdk/src/intTest/java/momento/sdk/retry/FixedDelayRetryStrategyIntegTest.java
```diff
@@ -91,7 +91,7 @@ void testNonRetryEligibleApi_shouldMakeNoAttempts_WhenFullNetworkOutage() throws
           assertThat(
                   testRetryMetricsCollector.getTotalRetryCount(
                       cacheName, MomentoRpcMethod.INCREMENT))
-              .isGreaterThanOrEqualTo(0);
+              .isEqualTo(0);
```
rishtigupta, line 94: Thnx!

momento-sdk/src/intTest/java/momento/sdk/retry/FixedDelayRetryStrategyIntegTest.java
```diff
@@ -66,7 +66,7 @@ void testRetryEligibleApi_shouldMakeMaxAttempts_WhenFullNetworkOutage() throws E
 
   @Test
   void testNonRetryEligibleApi_shouldMakeNoAttempts_WhenFullNetworkOutage() throws Exception {
-    RetryEligibilityStrategy eligibilityStrategy = (status, methodName) -> true;
+    RetryEligibilityStrategy eligibilityStrategy = (status, methodName) -> false;
```
rishtigupta, line 69: There is another test for non-retry eligible api with temporary outage. Can we make the change there too?

momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
```diff
@@ -115,6 +113,12 @@ public void onClose(Status status, Metadata trailers) {
                   return;
                 }
 
+                // If the deadline is expired, we don't want to retry
+                if (callOptions.getDeadline() != null && callOptions.getDeadline().isExpired()) {
```
rishtigupta, line 117: Where is the deadline being set in call options?

Does this mean that the following test for FixedDelayRetryStrategy to throw TIMEOUT_ERROR instead of a MISS?

```
@Test
  void
      testRetryEligibleApi_shouldMakeNoRetries_WhenTestConfigDelayMillisIsGreaterThanClientMaxDelayMillis()
          throws Exception {
    long maxDelayMillis = 1000L;

    FixedDelayRetryStrategy retryStrategy =
        new FixedDelayRetryStrategy(maxAttempts, delayMillis, maxDelayMillis);

    TestRetryMetricsMiddlewareArgs testRetryMetricsMiddlewareArgs =
        new TestRetryMetricsMiddlewareArgs.Builder(
                logger, testRetryMetricsCollector, UUID.randomUUID().toString())
            .delayRpcList(Collections.singletonList(MomentoRpcMethod.GET.getRequestName()))
            .delayMillis(1500) // greater than client max delay millis
            .build();

    withCacheAndCacheClient(
        config -> config.withRetryStrategy(retryStrategy).withTimeout(CLIENT_TIMEOUT_MILLIS),
        testRetryMetricsMiddlewareArgs,
        (cacheClient, cacheName) -> {
          assertThat(cacheClient.get(cacheName, "key"))
              .succeedsWithin(Duration.ofSeconds(10))
              .isInstanceOf(GetResponse.Miss.class);

          assertThat(testRetryMetricsCollector.getTotalRetryCount(cacheName, MomentoRpcMethod.GET))
              .isEqualTo(0);
        });
  }
```

nand4011: The deadline is set when the stub for the call is created. That test returns a miss because it doesn't retry, not because it hits the deadline, so it won't return the timeout error. The deadline for this test is set to 5000 ms, so it should never hit it.

momento-sdk/src/intTest/java/momento/sdk/retry/FixedDelayRetryStrategyIntegTest.java
nand4011: I might remove that test, since it doesn't matter if the network outage is temporary or full if the api isn't eligible, as it will never retry to start with.

nand4011: I thought it was testing an eligible api. I can change the name of the test if that was your intention.

rishtigupta: You're right. I am fine with removing the test.

rishtigupta (approved):
:shipit:

PR #433, opened 2025-03-13
chore: add Momento local tests for topics

nand4011:
Add TopicClientLocalTest containing tests for topics using Momento local. There are tests for subscribe retries and for the publish deadline.

Change the wait between retries for a subscription from 5 seconds to 0.5 seconds to bring it in line with the JS implementation.

Call the onError callback when a subscription receives an error that isn't `unavailable`.

Add middleware to the topic client.

Rename some of the retry test utility classes after Momento local.

Make the test metrics collector optional in the Momento local middleware.

Change the RPC list arguments from strings to MomentoRpcMethods for type safety.

Update the Momento local middleware args to return Optionals for optional fields.


momento-sdk/src/intTest/java/momento/sdk/retry/TestRetryMetricsMiddlewareTest.java
```diff
@@ -4,17 +4,18 @@
...
 import momento.sdk.exceptions.MomentoErrorCode;
 import momento.sdk.exceptions.MomentoErrorCodeMetadataConverter;
-import momento.sdk.retry.utils.TestRetryMetricsMiddlewareArgs;
-import momento.sdk.retry.utils.TestRetryMetricsMiddlewareRequestHandler;
+import momento.sdk.retry.utils.MomentoLocalMiddlewareArgs;
+import momento.sdk.retry.utils.MomentoLocalMiddlewareRequestHandler;
 import org.junit.jupiter.api.Test;
 
-public class TestRetryMetricsMiddlewareTest extends BaseCacheRetryTestClass {
+public class TestRetryMetricsMiddlewareTest extends BaseMomentoLocalTestClass {
```
rishtigupta, line 18: Rename the test class to `MomentoLocalMetricsMiddlewareTest`?

momento-sdk/src/main/java/momento/sdk/ScsTopicGrpcStubsManager.java
```diff
@@ -36,13 +41,22 @@ final class ScsTopicGrpcStubsManager implements Closeable {
...
     // set additional channel options (message size, keepalive, auth, etc)
     GrpcChannelOptions.applyGrpcConfigurationToChannelBuilder(
-        configuration.getTransportStrategy().getGrpcConfiguration(), channelBuilder);
+        configuration.getTransportStrategy().getGrpcConfiguration(),
+        channelBuilder,
+        credentialProvider.isEndpointSecure());
 
     final List<ClientInterceptor> clientInterceptors = new ArrayList<>();
+
+    final List<Middleware> middlewares = configuration.getMiddlewares();
```
rishtigupta, line 55: I remember javascript sdk also keeps a track of context for middlewares. Can you remind me why is it necessary?

momento-sdk/src/main/java/momento/sdk/config/TopicConfiguration.java
```diff
@@ -38,4 +55,25 @@ public TopicConfiguration withTimeout(final Duration timeout) {
...
+    newMiddlewares.add(middleware);
+    return new TopicConfiguration(this.transportStrategy, newMiddlewares, this.logger);
+  }
+
+  /**
+   * Get the middleware to be applied to the request.
+   *
+   * @return The middleware
+   */
+  public List<Middleware> getMiddlewares() {
```
rishtigupta, line 76: Java SDK does not have an `addMiddleware` helper function, unlike the [JS SDK](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/config/configuration.ts#L112). For consistency across SDKs, we might consider adding it to both the cache and topic clients in the Java SDK. However, this is not a blocker for this PR.

momento-sdk/src/intTest/java/momento/sdk/retry/BaseMomentoLocalTestClass.java
```diff
@@ -105,8 +108,50 @@ public static void withCacheAndCacheClient(
...
+            .build();
+
+    try {
+      if (cacheClient.createCache(cacheName).join() instanceof CacheCreateResponse.Error) {
+        throw new RuntimeException("Failed to create cache: " + cacheName);
+      }
+      testCallback.run(topicClient, cacheName);
+    } finally {
+      topicClient.close();
+      cacheClient.deleteCache(cacheName).join(); // Cleanup cache
```
rishtigupta, line 143: Is this necessary? momento-local doesn't keep a track of caches once the connection is severed?

rishtigupta: It is a good practice anyway, not complaining on cleaning up the resources 😆

momento-sdk/src/intTest/java/momento/sdk/retry/TopicClientLocalTest.java
```diff
@@ -0,0 +1,207 @@
...
+                    final int numRetries = connectionLostCounter.get();
+                    assertThat(numRetries).isGreaterThan(0);
+
+                    // wait for the connection to be restored, then send more messages to trigger
+                    // the error again
+                    connectionRestoredSemaphore.acquire();
+                    publishMessages(topicClient, cacheName, topicName, streamErrorMessageLimit);
+                    connectionLostSemaphore.acquire();
+
+                    assertThat(connectionLostCounter).hasValueGreaterThan(numRetries);
```
rishtigupta, line 106: Why are we not asserting the exact no. of retries that we expect?

anitarua (commented):
lgtm! I'm not as familiar with Java though so will leave final review to Rishti

momento-sdk/src/main/java/momento/sdk/config/TopicConfiguration.java
nand4011: It's called `withMiddleware` in Java. I copied the name from the [same method](https://github.com/momentohq/client-sdk-java/blob/main/momento-sdk/src/main/java/momento/sdk/config/Configuration.java#L130) in the cache configuration.

momento-sdk/src/intTest/java/momento/sdk/retry/BaseMomentoLocalTestClass.java
nand4011: I don't think it is necessary. the cache client version of this method has it, so I left it in when I copied that. I'll remove both of them.

momento-sdk/src/intTest/java/momento/sdk/retry/TopicClientLocalTest.java
nand4011: I figured that, since there is a timing element to the test, asserting that the number of retries is higher than last time we checked was less likely to cause transient errors. I can make it more exact if you want.

momento-sdk/src/main/java/momento/sdk/ScsTopicGrpcStubsManager.java
nand4011: It looks like a way to pass extra information into a middleware when setting it up. It is only used by the experimental metrics middleware to pass in a per client id. I don't think we need to preemptively add a similar thing to Java.

rishtigupta (approved):
:shipit:

PR #434, opened 2025-03-18
feat: add fixed timeout retry strategy

rishtigupta:
## PR Description:
- Add `FixedTimeoutRetryStrategy` 

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1185

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
+++ b/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
@@ -5,6 +5,7 @@ import io.grpc.Channel;
 import io.grpc.ClientCall;
 import io.grpc.ClientInterceptor;
 import io.grpc.Context;
+import io.grpc.Deadline;
 import io.grpc.ForwardingClientCallListener;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
@@ -16,6 +17,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
+import momento.sdk.retry.FixedTimeoutRetryStrategy;
 import momento.sdk.retry.RetryEligibilityStrategy;
 import momento.sdk.retry.RetryStrategy;
 import momento.sdk.retry.RetryingClientCall;
@@ -80,6 +82,9 @@ final class RetryClientInterceptor implements ClientInterceptor {
 
     return new RetryingClientCall<ReqT, RespT>(channel.newCall(method, callOptions)) {
       private int attemptNumber = 0;
+      private long cumulativeDelayMillis = 0; // cumulative delay for all retry attempts
+      private long lastAttemptTimeMillis =
+          System.currentTimeMillis(); // track time of the last attempt
       @Nullable private Future<?> future = null;
 
       @Override
@@ -113,8 +118,11 @@ final class RetryClientInterceptor implements ClientInterceptor {
                   return;
                 }
 
+                Deadline overallDeadline = callOptions.getDeadline();
+                long currentTimeMillis = System.currentTimeMillis();
+
                 // If the deadline is expired, we don't want to retry
-                if (callOptions.getDeadline() != null && callOptions.getDeadline().isExpired()) {
+                if (overallDeadline != null && overallDeadline.isExpired()) {
                   super.onClose(Status.DEADLINE_EXCEEDED, trailers);
                   return;
                 }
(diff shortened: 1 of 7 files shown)


momento-sdk/src/main/java/momento/sdk/retry/FixedTimeoutRetryStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * <p>The `FixedTimeoutRetryStrategy` implements the {@link RetryStrategy} interface and provides a
+ * fixed delay interval between each retry attempt. This delay interval includes a jitter factor to
+ * add randomness to the retry intervals, preventing retries from occurring at regular, predictable
+ * intervals.
+ *
+ * <p>Additionally, the strategy enforces a maximum cumulative delay (specified by
+ * `responseDataReceivedTimeoutMillis`). If the cumulative delay from all retry attempts exceeds
+ * this maximum value, no further retries will be performed, and the strategy will return an empty
+ * optional.
```
anitarua, line 22: maybe copypasta, JS comment for responseDatareceivedTimeoutMillis [here](https://github.com/momentohq/client-sdk-javascript/blob/f1b3b103f4f7748619ca1b3cfd7ff21a7fff8553/packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts#L16)

momento-sdk/src/main/java/momento/sdk/retry/FixedTimeoutRetryStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Constructs a `FixedTimeoutRetryStrategy` with the provided parameters.
+   *
+   * @param eligibilityStrategy a strategy that determines if the gRPC status code and methods are
+   *     eligible or safe to retry. If null, a default strategy is used.
+   * @param retryDelayIntervalMillis The delay in milliseconds for each retry attempt. If null,
+   *     default value of 100 milliseconds is used.
+   * @param responseDataReceivedTimeoutMillis The maximum cumulative delay in milliseconds that is
+   *     allowed for all retry attempts combined. If the cumulative delay exceeds this value, no
+   *     more retries will be performed, and the strategy will return an empty optional. If null,
```
anitarua, line 49: maybe more copypasta. I think there's no "maximum cumulative delay" concept in this strategy

momento-sdk/src/test/java/momento/sdk/retry/FixedTimeoutRetryStrategyTest.java
```diff
@@ -0,0 +1,59 @@
...
+  @Test
+  void testGetTimeout_SuccessfulRetry() {
+    long retryIntervalDelayMillis = 100;
+    long responseDataReceivedTimeoutMillis = 200;
+    FixedTimeoutRetryStrategy retryStrategy =
+        new FixedTimeoutRetryStrategy(
+            eligibilityStrategy, retryIntervalDelayMillis, responseDataReceivedTimeoutMillis);
+    Optional<Duration> delay = retryStrategy.determineWhenToRetry(status, methodDescriptor, 1);
+    assertTrue(delay.isPresent());
+    assertTrue(delay.get().toMillis() <= responseDataReceivedTimeoutMillis);
```
anitarua, line 45: I think the test should assert that if eligible, determineWhenToRetry will return retryIntervalDelayMillis +/- some jitter

momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
```diff
@@ -80,6 +82,9 @@ public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
 
     return new RetryingClientCall<ReqT, RespT>(channel.newCall(method, callOptions)) {
       private int attemptNumber = 0;
+      private long cumulativeDelayMillis = 0; // cumulative delay for all retry attempts
```
anitarua, line 85: I don't think there should be any concept of cumulative delay when using this strategy.

The JS fixed timeout strategy causes the retry interceptor to [override the deadline](https://github.com/momentohq/client-sdk-javascript/blob/f1b3b103f4f7748619ca1b3cfd7ff21a7fff8553/packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts#L66-L68) to be responseDataReceivedTimeoutMillis from current time, as long as we have not reached the overall deadline set by the client timeout millis config.

The flow of the fixed timeout strategy is:

- client sets ClientTimeoutMillis, this determines the overall deadline for a unary request
- initial request attempt comes back with an error that is eligible for retry
- as long as we have not yet passed the overall deadline set by ClientTimeoutMillis, we keep retrying the request like so:
    - determineWhenToRetry will tell us how long to delay our retry attempt (e.g. send the request 100ms +/- jitter from now)
    - retry interceptor overrides the retry attempt's grpc deadline to responseDataReceivedTimeoutMillis from now, meaning if we do not receive a response within responseDataReceivedTimeoutMillis, we should get a timeout error for our retry attempt 
    - if a retry attempt fails, we repeat the process (determine next retry time, override deadline again using responseDataReceivedTimeoutMillis)

momento-sdk/src/main/java/momento/sdk/retry/FixedTimeoutRetryStrategy.java
rishtigupta: Just chatgpt being chatgpt 🤣

momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java
```diff
@@ -140,7 +141,44 @@ public void onClose(Status status, Metadata trailers) {
...
-                    Context.current().wrap(() -> retry(channel.newCall(method, callOptions)));
+                    Context.current()
+                        .wrap(
+                            () -> {
+                              Optional<Long> responseTimeout =
+                                  retryStrategy.getResponseDataReceivedTimeoutMillis();
+
+                              // Check if the overall deadline is set and if the retry deadline
+                              // exceeds it
+                              if (responseTimeout.isPresent()) {
```
rishtigupta, line 152: Currently, I added this logic to be applied only for timeout strategy, but I think it can be extended for any strategy. wdyt?
I am happy to discuss it on a call as well.

anitarua: since it's implemented in the retry interceptor like in the js version, wouldn't this apply to all retry strategies that provide a value for `getResponseDataReceivedTimeoutMillis()` anyway? (which is fine, I think the same in JS)

anitarua (approved):
lgtm!

PR #435, opened 2025-03-19
chore: improve subscription reconnection logic

nand4011:
Move the topic subscription reconnection logic into a retry strategy and eligibility strategy that can be configured by the client.

Add the subscription retry strategy to TopicConfiguration. Add new constructors that take the retry strategy, and new ones that don't take a logger, which is only used inside the TopicConfiguration itself.

Prevent potential race conditions that could occur when unsubscribing from a connection while it is reconnecting by checking if the user called unsubscribe before reconnecting, and by making the variables used in reconnection atomic.

Call close in SubscriptionWrapper when a stream ends to stop a subscription from leaking threads from the retry executor service. We may want to share a thread pool between the subscriptions instead of making a new one per subscription, so that we don't need to make a new one each time. We could also use it to execute the callbacks, which are currently executed in the gRPC thread pool and could cause backpressure if they are long-running.

Add empty default implementations for the test methods in IScsTopicConnection to clean up its use in the topic client.

Remove SendSubscribeOptions, since we already have a class that contains the callbacks, and because it had an unused subscription object.

Get rid of CancelableClientCallStreamObserver's ClientCallStreamObserver extension, because that class is meant for observing requests, not responses, and we only need a simple cancel method.

Make TopicClientLocalTest's unrecoverable error 'not found', because the previous error is now recoverable.


momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -5,104 +5,120 @@
...
    *
    * @return A CompletableFuture representing the asynchronous execution of the internal
    *     subscription logic with retry mechanism.
    */
-  public CompletableFuture<Void> subscribeWithRetry() {
-    CompletableFuture<Void> future = new CompletableFuture<>();
-    subscribeWithRetryInternal(future);
-    return future;
-  }
+  CompletableFuture<Void> subscribeWithRetry() {
```
nand4011, line 70: I combined this with the internal method because the public method didn't do anything other than creating the future, and this way we don't have to give an invalid future to the internal method if we retry. I also made it package private so we can call if for unit tests without letting end users have access.

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -5,104 +5,120 @@
...
+    final CompletableFuture<Void> future = new CompletableFuture<>();
+
+    if (!isSubscribed.get()) {
+      completeExceptionally(
+          future, new UnknownException("Cannot subscribe to an unsubscribed subscription"));
```
nand4011, line 75: I wasn't sure what type of error code to use here.

momento-sdk/src/main/java/momento/sdk/retry/FixedDelaySubscriptionRetryStrategy.java
```diff
@@ -0,0 +1,32 @@
...
+    this.eligibilityStrategy = eligibilityStrategy;
+    this.retryDelay = retryDelay;
+  }
+
+  public FixedDelaySubscriptionRetryStrategy(Duration retryDelay) {
+    this(new DefaultSubscriptionRetryEligibilityStrategy(), retryDelay);
+  }
+
+  @Override
+  public Optional<Duration> determineWhenToRetry(Throwable error) {
```
nand4011, line 25: Do we need anything other than the error type? I'm using the error instead of the status so I can move some more retry logic into here. The gRPC method doesn't matter since this is only used for subscriptions. We might care about attempt number if we want a retry strategy that has an increasing delay between attempts, but the only implementation has a fixed delay for now.

Also, do we want to use the word 'retry' in the naming? It felt more natural in the documentation to use 'reconnect' to distinguish between reconnecting an existing subscription and retrying the call that made the subscription.

malandis: Shouldn't the retry eligibility be a separate concern vs `determineWhenToRetry`? Ie `determineWhenToRetry` should be invoked if and only if the call is eligible for retry. Otherwise anyone who implements a `RetryStrategy` will have to include the `isEligibleForRetry` test in their implementation.

nand4011: The cache client calls don't call the eligibility strategy independently from the retry strategy; it's always called by the retry strategy.

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
```diff
@@ -5,104 +5,120 @@
...
-  private boolean firstMessage = true;
-  private boolean isConnectionLost = false;
+  private final String cacheName;
+  private final String topicName;
+  private final ISubscriptionCallbacks callbacks;
+  private final SubscriptionState subscriptionState;
+  private final SubscriptionRetryStrategy retryStrategy;
+  private final AtomicBoolean firstMessage = new AtomicBoolean(true);
+  private final AtomicBoolean isConnectionLost = new AtomicBoolean(false);
+  private final AtomicBoolean isSubscribed = new AtomicBoolean(true);
```
anitarua, line 39: is this used instead of SubscriptionState's `private boolean isSubscribed`?

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
anitarua: is there a client sdk error in this sdk? otherwise unknown seems reasonable enough to me

momento-sdk/src/main/java/momento/sdk/retry/FixedDelaySubscriptionRetryStrategy.java
anitarua: Keeping track of attempt number could be nice, especially if printing in debug logs.

And since Java would be the first one with the subscription-specific retry strategy, I think either naming scheme would be fine.

momento-sdk/src/main/java/momento/sdk/SubscriptionWrapper.java
nand4011: I'll look into combining those.

momento-sdk/src/main/java/momento/sdk/retry/DefaultSubscriptionRetryEligibilityStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+      LoggerFactory.getLogger(DefaultSubscriptionRetryEligibilityStrategy.class);
+
+  private static final Set<Status.Code> NON_RETRYABLE_STATUS_CODES;
+
+  static {
+    final Set<Status.Code> nonRetryableStatusCodes = new HashSet<>();
+    nonRetryableStatusCodes.add(Status.Code.PERMISSION_DENIED);
+    nonRetryableStatusCodes.add(Status.Code.UNAUTHENTICATED);
+    nonRetryableStatusCodes.add(Status.Code.CANCELLED);
+    nonRetryableStatusCodes.add(Status.Code.NOT_FOUND);
```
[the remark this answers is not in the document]
nand4011: That is for when a cache is deleted mid-stream. I can make an integration test to show that that is the error returned when that happens.

anitarua:
Rebased onto main and added a commit for gracefully ending subscriptions when unsubscribe is called (should call onCompleted instead of onError). Will use this as the new base for subscriptions bookkeeping changes (https://github.com/momentohq/client-sdk-java/pull/448)

malandis (approved):

PR #439, opened 2025-03-28
fix: minor fixes for retry tests

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1207

Assigns the Min(retry deadline, overall deadline) on each retry attempt to ensure we don't exceed the client's overall deadline.

Ended up removing the check for "if retry delay ms will cause retry attempt to exceed client timeout, then don't retry" since that's not included in any other implementations so far. Adjusted one test case to accommodate--depending on jitter, the test may make one retry attempt that will timeout before it can complete.


rishtigupta (dismissed):
:shipit:

rishtigupta (approved):
:shipit:

PR #442, opened 2025-04-08
feat: add configurations for topics with separate unary and stream channels

rishtigupta:
## PR Description:
This PR introduces support for independently configuring the number of unary and streaming gRPC channels for the Topics client. It includes the following key changes:

- Added configuration support for specifying distinct values for:
    - `numStreamGrpcChannels` (default: 4)
    - `numUnaryGrpcChannels` (default: 4)
- Refactored `ScsTopicGrpcStubsManager` to create and manage separate pools of stubs for unary and streaming operations. 

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1155


momento-sdk/src/main/java/momento/sdk/config/transport/topics/TopicsGrpcConfiguration.java
```diff
@@ -0,0 +1,370 @@
...
+        keepAliveTime);
+  }
+
+  /**
+   * @deprecated This is a no-op. Use {@link #getNumStreamGrpcChannels()} (int)} and {@link
+   *     #getNumUnaryGrpcChannels()} (int)} instead.
+   * @return the minimum number of gRPC channels.
+   */
+  @Override
+  public int getMinNumGrpcChannels() {
```
nand4011, line 129: Do we need this? We are already breaking backwards compatibility by replacing the TransportStrategy in TopicConfiguration with a TopicsTransportStrategy.

momento-sdk/src/main/java/momento/sdk/config/transport/topics/TopicsGrpcConfiguration.java
```diff
@@ -0,0 +1,370 @@
...
+   * @param keepAliveWithoutCalls Whether to send keepalive pings without any active calls.
+   * @param keepAliveTimeout The time in milliseconds to wait for a keepalive ping response before
+   *     considering the connection dead.
+   * @param keepAliveTime The time in milliseconds to wait between keepalive pings.
+   */
+  public TopicsGrpcConfiguration(
+      @Nonnull Duration deadline,
+      int numStreamGrpcChannels,
+      int numUnaryGrpcChannels,
+      Optional<Integer> maxMessageSize,
```
nand4011, line 56: Optionals in Java should only be used as return types, since you could still pass null instead of an Optional here. I don't think we need this constructor if we have the one below. I like the Optional usage for the get methods below though.

momento-sdk/src/main/java/momento/sdk/config/transport/topics/TopicsTransportStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+  TopicsTransportStrategy withGrpcConfiguration(TopicsGrpcConfiguration grpcConfiguration);
+
+  /**
+   * The maximum number of concurrent requests that the Momento client will allow onto the wire at a
+   * given time.
+   *
+   * @return the max requests, or null if there is no maximum
+   */
+  @Nullable
+  Integer getMaxConcurrentRequests();
```
nand4011, line 30: I would make this return an `Optional<Integer>` to match the use of optional in TopicsGrpcConfiguration.

momento-sdk/src/main/java/momento/sdk/ScsTopicGrpcStubsManager.java
```diff
@@ -62,9 +92,16 @@ private static ManagedChannel setupConnection(
...
   }
 
-  /** Returns a pubsub stub. */
-  PubsubGrpc.PubsubStub getStub() {
-    return stub;
+  /** Round-robin publish stub. */
+  PubsubGrpc.PubsubStub getNextUnaryStub() {
+    return unaryStubs
+        .get(unaryIndex.getAndIncrement() % this.numUnaryGrpcChannels)
+        .withDeadlineAfter(deadline.toMillis(), TimeUnit.MILLISECONDS);
```
nand4011, line 99: I'm glad we can move this into the stubs manager.

nand4011 (commented):
Have we confirmed that we're alright with breaking backwards compatibility to make a TopicsTransportStrategy and TopicsGrpcConfiguration? Another option would be to add a getNumStreamingChannels method to GrpcConfigration and always set it to 0 for the cache client.

anitarua (dismissed):
lgtm! but may want @nand4011 to give a final pass

momento-sdk/src/main/java/momento/sdk/config/transport/GrpcConfiguration.java
```diff
@@ -13,6 +13,8 @@ public class GrpcConfiguration implements IGrpcConfiguration {
 
   private final @Nonnull Duration deadline;
   private final int minNumGrpcChannels;
```
nand4011, line 15: What does this mean now that we have numUnaryGrpcChannels? It effectively was the number of grpc channels to use before.

momento-sdk/src/main/java/momento/sdk/config/transport/GrpcConfiguration.java
```diff
@@ -48,13 +54,17 @@ public GrpcConfiguration(@Nonnull Duration deadline) {
   public GrpcConfiguration(
```
nand4011, line 54: You will need a constructor that doesn't take numStreamGrpcChannels and numUnaryGrpcChannels for backwards compatibility.

momento-sdk/src/main/java/momento/sdk/config/transport/GrpcConfiguration.java
nand4011: The name minNumGrpcChannels isn't great, but why not treat that as numUnaryGrpcChannels and then only add a numStreamGrpcChannels?

rishtigupta: To maintain consistency with the JS SDK, I introduced separate configurations for both unary and stream gRPC channels. The existing `minNumGrpcChannels` name suggests that the SDK might dynamically increase the number of channels at runtime, which isn't the case. To avoid this confusion and make the behavior clearer to users, I opted for explicit `numUnaryGrpcChannels` and `numStreamGrpcChannels` settings.

nand4011: We should change the other stubs managers to use the new setting then, since a customer could set `numUnaryGrpcChannels` and expect that to affect the number of channels used by a cache client.

anitarua (dismissed):
lgtm!

nand4011 (approved):

PR #448, opened 2025-05-21
fix: add bookkeeping for number of active subscriptions

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1231

- Found that the java sdk does not keep track of the number of concurrent subscriptions while auditing the implementations of separate grpc channels for publish and subscribe. 
    - This PR adds the necessary bookkeeping so that the TopicClient can reject new subscribe requests if all grpc channels are currently occupied with active streams. 
    - IIRC, Java has a multi-threaded async runtime, meaning multiple threads could be simultaneously handling multiple subscribe requests. Need to ensure the bookkeeping takes this into account so bursts of subscribe/unsubscribe requests are handled properly. 

- Also added a new retry test file to verify that the TopicClient does not allow >100 streams on a single grpc channel and that it is capable of handling bursts of subscribe and unsubscribe requests. 

Note: this PR is based on top of Nate's [PR](https://github.com/momentohq/client-sdk-java/pull/435) for other subscriptions improvements


momento-sdk/src/main/java/momento/sdk/ScsTopicGrpcStubsManager.java
```diff
@@ -100,8 +135,30 @@ PubsubGrpc.PubsubStub getNextUnaryStub() {
...
 
   /** Round-robin subscribe stub. */
-  PubsubGrpc.PubsubStub getNextStreamStub() {
-    return streamStubs.get(streamIndex.getAndIncrement() % this.numStreamGrpcChannels);
+  StreamStubWithCount getNextStreamStub() {
+    // First check if there's available capacity on any of the stubs
+    int totalActiveSubscriptions = 0;
+    for (StreamStubWithCount stubWithCount : streamStubs) {
+      totalActiveSubscriptions += stubWithCount.getCount();
+    }
```
malandis, line 143: Do we need to tally this up given we check each channel below individually for capacity?

momento-sdk/src/main/java/momento/sdk/ScsTopicGrpcStubsManager.java
```diff
@@ -100,8 +135,30 @@ PubsubGrpc.PubsubStub getNextUnaryStub() {
...
+    if (totalActiveSubscriptions < this.maximumActiveSubscriptions) {
+      // Try to get a client with capacity for another subscription.
+      // Allow up to maximumActiveSubscriptions attempts.
+      for (int i = 0; i < this.maximumActiveSubscriptions; i++) {
+        // Round-robin to the next stub
+        final StreamStubWithCount stubWithCount =
+            streamStubs.get(streamIndex.getAndIncrement() % this.numStreamGrpcChannels);
+        if (stubWithCount.getCount() < 100) {
+          stubWithCount.incrementCount();
```
malandis, line 152: There's a TOCTOU bug here. The stubWithCount should have an acquire or fail method that atomically acquires a resource (tallies one to the count) ro else fails. When we fail we try a different channel.

momento-sdk/src/main/java/momento/sdk/ScsTopicGrpcStubsManager.java
```diff
@@ -100,8 +142,27 @@ PubsubGrpc.PubsubStub getNextUnaryStub() {
...
+      final StreamStubWithCount stubWithCount =
+          streamStubs.get(streamIndex.getAndIncrement() % this.numStreamGrpcChannels);
+      try {
+        stubWithCount.acquireStubOrThrow();
+        return stubWithCount;
+      } catch (ClientSdkException e) {
+        // If the stub is at capacity, continue to the next one.
+        continue;
+      }
+    }
```
malandis, line 160: 👍

malandis (approved):

PR #450, opened 2025-06-17
feat: add WithMaxSubscriptions config backed by stream vs unary grpc connection pools

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1164

Opted for minimal config changes to maintain backwards compatibility.

- Added `WithMaxSubscriptions` and an `IsNumStreamChannelsDynamic` flag to the `TopicConfiguration`
- Topics configs will default to using a static pool of stream grpc channels for subscriptions (defaults to 4 for both unary and stream)
- Using WithMaxSubscriptions will tell the topic client to use a dynamic pool instead (starts from 1 channel, grows to support the specified maxSubscriptions)

Also added more tests to the subscription initialization tests (part of the retry tests) to exercise the dynamic channel creation. The main difference between the static and dynamic pools are:

- `StaticStreamGrpcConnectionPool` uses a regular `java.util.List` that is initialized once in the constructor and is never modified.
- `DynamicStreamGrpcConnectionPool` uses `java.util.concurrent.CopyOnWriteArrayList` and `AtomicInteger` to be able to dynamically add channels as needed up until the maximum number of channels has been reached.

### DIFF
--- a/momento-sdk/src/main/java/momento/sdk/DynamicStreamGrpcConnectionPool.java
+++ b/momento-sdk/src/main/java/momento/sdk/DynamicStreamGrpcConnectionPool.java
@@ -53,8 +53,10 @@ public class DynamicStreamGrpcConnectionPool implements StreamTopicGrpcConnectio
             .collect(Collectors.toCollection(CopyOnWriteArrayList::new));
   }
 
-  // Multiple threads could get to the point of seeing currentNumActiveStreams == currentMaxConcurrentStreams,
-  // but we need to ensure only one thread will add a new channel at a time so that we don't exceed the max number of channels.
+  // Multiple threads could get to the point of seeing currentNumActiveStreams ==
+  // currentMaxConcurrentStreams,
+  // but we need to ensure only one thread will add a new channel at a time so that we don't exceed
+  // the max number of channels.
   private void addNewChannel() {
     final int updatedCount = this.currentNumStreamGrpcChannels.incrementAndGet();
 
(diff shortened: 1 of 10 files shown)


momento-sdk/src/main/java/momento/sdk/config/TopicConfiguration.java
```diff
@@ -147,4 +174,37 @@ public TopicConfiguration withMiddleware(@Nonnull final Middleware middleware) {
...
+   * accommodate, but will start with one channel which supports 100 subscriptions.
+   *
+   * @param maxSubscriptions The new maximum number of active subscriptions.
+   * @return a new TopicConfiguration with the updated maximum number of active subscriptions.
+   */
+  public TopicConfiguration withMaxSubscriptions(int maxSubscriptions) {
+    if (maxSubscriptions < 0) {
+      throw new InvalidArgumentException("maxSubscriptions must be greater than 0");
+    }
+    final int maxStreamChannels = (int) Math.ceil(maxSubscriptions / 100.0);
```
malandis: Let's extract this to a variable at least, eg numStreamsPerChannel.

momento-sdk/src/main/java/momento/sdk/StreamTopicGrpcConnectionPool.java
```diff
@@ -0,0 +1,49 @@
...
+  int incrementCount() {
+    return count.incrementAndGet();
+  }
+
+  int decrementCount() {
+    return count.decrementAndGet();
+  }
+
+  void acquireStubOrThrow() throws ClientSdkException {
+    if (count.incrementAndGet() <= 100) {
```
malandis: For maintainability we should extract or externalizae these magic numbers.

momento-sdk/src/main/java/momento/sdk/DynamicStreamGrpcConnectionPool.java
```diff
@@ -0,0 +1,120 @@
...
+    // Otherwise return an error if no stubs have capacity.
+    throw new ClientSdkException(
+        MomentoErrorCode.CLIENT_RESOURCE_EXHAUSTED,
+        "Maximum number of active subscriptions reached");
+  }
+
+  @Override
+  public void close() {
+    streamChannels.forEach(ManagedChannel::shutdown);
+  }
```
malandis, line 123: This class should implement `Closeable` since that is a convention.

malandis (approved):

PR #460, opened 2026-09-14
feat!: require protobuf 4, bump grpc, and remove store clients

tylerburdsall:
Breaking changes:
* `store` client is removed
* `protobuf-java@4.x` is now required at runtime

Addresses some open CVEs


PDXKimani (commented):
As I believe this is pretty breaking, do we need to do anything in particular to cause a new major revision/version of this SDK on release?

malandis:
> As I believe this is pretty breaking, do we need to do anything in particular to cause a new major revision/version of this SDK on release?

Why is this a breaking change?

tylerburdsall:
> > As I believe this is pretty breaking, do we need to do anything in particular to cause a new major revision/version of this SDK on release?
> 
> 
> 
> Why is this a breaking change?

@malandis see PR description

malandis (approved):

